PR #5, opened 2018-01-23
Feature/store version and converter

jcwinkler:
A tiny pr - to warm up on github :-)

The hard coded strings "pdflatex" and "pstoedit" will be replaced later by class methods returning the required information (see detailed commit message).

### DIFF
--- a/textext.py
+++ b/textext.py
@@ -307,7 +307,7 @@ class TexText(inkex.Effect):
             return
 
         # -- Store textext attributes
-        new_node.attrib['{%s}textextver' % TEXTEXT_NS] = __version__.encode('string-escape')
+        new_node.attrib['{%s}version' % TEXTEXT_NS] = __version__.encode('string-escape')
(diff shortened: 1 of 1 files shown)


leberwurstsaft:
Very good idea. I'd argue to rename "textextver" to "version", since the namespace is already in there.

textext.py
```diff
@@ -307,6 +307,9 @@ def do_convert(self, text, preamble_file, scale_factor, converter_class, old_nod
             return
 
         # -- Store textext attributes
+        new_node.attrib['{%s}textextver' % TEXTEXT_NS] = __version__.encode('string-escape')
```
leberwurstsaft: Call it "version" instead of "textextver". The namespace is already about textext.

leberwurstsaft: The idea to store these attributes is great, btw.

jcwinkler: Right. Changed it.

PR #7, opened 2018-02-05
Feature/scale to uu

### LINKED ISSUE #1: Font size mismatch between Inkscape and TexText generated nodes
If the font in a LaTeX snippet is set explicitely to a certain size (e.g. 10pt) the text of the generated svg node has a much larger font size if the scale factor is 1.0. Expected behavior is that the font size is equal to the font size of text directly set in Inkscape.

Illustration

Originally reported on [Bitbucket as issue #55](https://bitbucket.org/pitgarbe/textext/issues/55/font-size-and-warning-message)

jcwinkler:
Resolves #1.

See detailed information on this fix in the discussion of issue #1.

(How is it possible to avoid that the changes of the zero length field name format issue #6 do show up here again? I needed them for development, so I cherry picked them, since I branched before they went into the master. ??? Thanks for any hints.)

### DIFF
--- a/textext.py
+++ b/textext.py
@@ -202,7 +202,12 @@ class TexText(inkex.Effect):
         # If we do not perform this operation nodes scaled by the user to his needs in the previous version
         # will be about three times smaller after insertion of the new node (if doc units are mm).
         if (old_node is not None) and ('{%s}version' % TEXTEXT_NS not in old_node.attrib.keys()):
-            current_scale *= self.uutounit(1, "pt")
+            try:
+                # Inkscape > 0.48
+                current_scale *= self.uutounit(1, "pt")
+            except AttributeError:
+                # Inkscape <= 0.48
+                current_scale *= inkex.uutounit(1, "pt")
 
         # Ask for TeX code
         if self.options.text is None:
--- a/textext.py
+++ b/textext.py
@@ -196,11 +196,9 @@ class TexText(inkex.Effect):
         # Find root element
         old_node, text, preamble_file, current_scale = self.get_old()
 
-        # Adapt scale factor of nodes created with previous versions of TexText
-        # This is only necessary for nodes created with TexText 0.7.0 and prior. These nodes
-        # do not have a version keyword.
-        # If we do not perform this operation nodes scaled by the user to his needs in the previous version
-        # will be about three times smaller after insertion of the new node (if doc units are mm).
+        # This is very important when re-editing nodes which have been created using TexText <= 0.7. It ensures that
+        # the scale factor which is displayed in the AskText dialog is adjusted in such a way that the size of the node
+        # is preserved when recompiling the LaTeX code.
         if (old_node is not None) and ('{%s}version' % TEXTEXT_NS not in old_node.attrib.keys()):
             try:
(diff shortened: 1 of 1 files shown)


leberwurstsaft:
To include changes from the originating branch, you could either rebase your branch or simply merge in the changes into yours. Rebasing is best done if you haven't yet pushed your branch, since it rewrites the history. Only if you're absolutely sure that no one has pulled the new branch yet, you could safely force push a rebased branch. Cherry picking is part of a (interactive) rebase and also an option IMO. In this case I'd say it doesn't matter all that much. Definitely better that manually copying a fix into the branch, as it should avoid conflicts.

That said, I just had a look at the branch network and it *looks like* there was another open branch by you. But it is already merged into master. From the network graph it looks like the work isn't in your current branch? That'd be odd. Can you confirm that the changes from the store_version_and_converter branch are in fact also in your current branch?

bildschirmfoto 2018-02-05 um 11 55 59

jcwinkler:
Yes, the changes from the store_version_and_converter branch are in my master. 

grafik

I keep my master in sync with this repo's master (according to https://help.github.com/articles/syncing-a-fork/) so the changes went into it via that way. But, of course after branching the scale_to_uu branch.  

Compared to the state of my master highlighting of the changes in asktext.py are correct. But compared to textext's original master I guess they are not correct, are they?

leberwurstsaft:
Right, it would have been strange anyway, since the new work depends on the one in the other branch. Still, even looking at the network graph in your fork https://github.com/jcwinkler/textext/network it paints a strange picture. Not sure what to make of that.

textext.py
```diff
@@ -279,7 +290,7 @@ def preview_convert(text, preamble_file, converter_class, image_setter):
             os.chdir(cwd)
             converter.finish()
 
-    def do_convert(self, text, preamble_file, scale_factor, converter_class, old_node):
+    def do_convert(self, text, preamble_file, scale_factor_from_user, converter_class, old_node):
```
leberwurstsaft: I was thinking about maybe renaming this parameter to user_scale_factor instead?

textext.py
```diff
@@ -369,7 +389,7 @@ def do_convert(self, text, preamble_file, scale_factor, converter_class, old_nod
             self.settings.set('preamble', '')
 
         if scale_factor is not None:
-            self.settings.set('scale', scale_factor)
+            self.settings.set('scale', scale_factor_from_user)
```
leberwurstsaft: Then this would also maybe be called user_scale_factor instead.  But thats just a matter of taste anyway.

jcwinkler:
OK, finally -knowing (or hoping...) that I am working alone on this branch- I did a `git rebase --force master
`and `git push --force` and now the pr includes what it is meant for,

textext.py
jcwinkler: Good idea, will change it.

textext.py
jcwinkler: see above

leberwurstsaft:
👍

leberwurstsaft:
I saw your replies to my comments after merging this PR. Sorry.

PR #12, opened 2018-03-28
Fixed: Failure on missing inkscape version key

### LINKED ISSUE #10: Error occurred while converting text from LaTeX to SVG
I have installed
Inkscape 0.92.1 r15371
ghostscript 9.06
pstoedit 3.62 

When type something in the textext gui and click "preview" I get a preview as expected. When I click "save" a window with the heading "Error occurred while converting text from LaTeX to SVG:" opens, containing the following message:
`u'{http://www.inkscape.org/namespaces/inkscape}version'
Traceback (most recent call last):
  File "/home/boris/.config/inkscape/extensions/asktext.py", line 491, in cb_ok
    self.callback(self.text, self.preamble_file, self.global_scale_factor)
  File "textext.py", line 224, in <lambda>
    old_node),
  File "textext.py", line 337, in do_convert
    self.document.getroot().attrib['{%s}version' % inkex.NSS["inkscape"]].split(' ')[0]).encode('string-escape')
  File "lxml.etree.pyx", line 2355, in lxml.etree._Attrib.__getitem__ (src/lxml/lxml.etree.c:62354)
KeyError: u'{http://www.inkscape.org/namespaces/inkscape}version'
`

jcwinkler:
Fixes #10 "Error occurred while converting text from LaTeX to SVG"

Unfortunately when a node processed by TexText comes from an Inkscape
document that has never been saved before no version attribute is
provided by Inkscape. The corresponding call to the key fails. In these
cases no version attribute is stored in the TexText node. It is not
ideal but currently I do not see any other solution. Completely ommiting
this piece of information is not an option in my opinion. We should grab
what can be grabbed.

### DIFF
--- a/textext.py
+++ b/textext.py
@@ -333,8 +333,13 @@ class TexText(inkex.Effect):
         new_node.attrib['{%s}text' % TEXTEXT_NS] = text.encode('string-escape')
         new_node.attrib['{%s}preamble' % TEXTEXT_NS] = preamble_file.encode('string-escape')
         new_node.attrib['{%s}scale' % TEXTEXT_NS] = str(user_scale_factor).encode('string-escape')
-        new_node.attrib['{%s}inkscapeversion' % TEXTEXT_NS] = (
-        self.document.getroot().attrib['{%s}version' % inkex.NSS["inkscape"]].split(' ')[0]).encode('string-escape')
+        try:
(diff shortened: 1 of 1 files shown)


leberwurstsaft:
Thanks!

Issue #13, opened 2018-03-28
Stretched object disappears after editing


sizmailov:
If object was stretched, it disappears from document after editing. 

Steps to reproduce:
1) Create regular object with textext
![1](https://user-images.githubusercontent.com/2975991/38026426-b73e9de8-3294-11e8-854b-67f7f4da7b77.png)

2) Stretch object in x or y direction
![2](https://user-images.githubusercontent.com/2975991/38026438-bd2a7efc-3294-11e8-97b9-1c41689b4e28.png)

3) Make an arbitrary change (or even no change) and save object.
![3](https://user-images.githubusercontent.com/2975991/38026442-c32fb966-3294-11e8-923d-d925096c37e1.png)

4) Object is gone
![4](https://user-images.githubusercontent.com/2975991/38026470-e13a6492-3294-11e8-8cb9-c364226b8098.png)

Note: skewed object does not disappear, and it's skew is preserved upon editing.

This issue seems to be connected to #11 

I realize that stretched text looks ugly and should not be used for 99.9% of time.

sizmailov:
I confirm, the object flew away, just as @jcwinkler said https://github.com/textext/textext/issues/11#issuecomment-376857091

Issue #11, opened 2018-03-27
Object scale is not preserved after editing


sizmailov:
Steps to reproduce:
1) Create formula "$\sin(x)$" with textext
2) Scale the resulting object to fit your layout
3) Change original formula, say "$\sin(x)+1$" 

Desired behavior: object scale same as at step 2
Actual behavior: object scale is reset

jcwinkler:
This is a misunderstanding of what the term "scale factor" means in TexText. "Scale" means that the length unit coming from the latex-svg node is scaled by the specified factor. 

So if you change the length or hight of the content -- as in the exmample -- then the size of the node will change, too, and the node is roughly placed in the middle of the olde one. Please see the attached figure. In all three nodes the sin(x) has the same size, but the recreated node with $\sin(x)+1$ larger in width while its hight is constant.

![grafik](https://user-images.githubusercontent.com/11866252/37969316-b26933ce-31d0-11e8-8c8c-2d28950134ef.png)

(There is a very small placing issue after rescaling/reediting. This will be fixed in the pdf2svg version of TexText)

sizmailov:
Let me be more clear about this issue. I'm perfectly fine with change of object width along with length of formula, it's not an issue. 

In the attached image I have three sets of figures and formulae, row by row: original formula, by-hand-scaled formula, edited-by-hand-scaled formula. My concern is edited-by-hand-scaled was reverted to original size and my hand-tuned scale is lost.

![1](https://user-images.githubusercontent.com/2975991/37972687-86eb181e-31e1-11e8-88db-673c43d58489.png)

jcwinkler:
This behaviour is intended. Although I understand that in your scenario this might by annoying.

Let me explain: When @leberwurstsaft implemented the scale factor feature some time ago we wanted to have the possibility to scale the svg-node resulting from pdflatex+pstoedit in such a way that it fits to the geometry of the other objects (boxes, lines, etc.) in the Inkscape drawing. On the other hand, from a typographic point of view, there should be only one or maybe two font sizes used in a single document. So, if one is of the opinion that a scale factor of e.g. 0.8 is suitable for the current Inkscape document all other nodes should also have a scale factor of 0.8. We think that this is more easily and intuitively realized by a number than by scaling each object manually by the mouse. Hence, the scale factor directly manipulates the transform-attribute of the svg-group and overwrites any changes done by the mouse in Inkscape.

[...]

PR #14, opened 2018-04-05
Fix bad positioning

sizmailov:
This patch fixes issues with position and scale of object that was scaled manually in inkscape and then re-edited with textext (see #11, #13). 

Now scaling of textext object in inkscape is coupled with "scale factor" in plugin window. This behavior match behavior of native text objects in inkscape, so this is definitely right way to handle it. 

Here are several cycles of `create -> edit in inkscape -> edit content in textext -> scale to 1.0 in textext` with different transformations in inkscape. 

g7354-9

Another improvement is addition of a control which allows to choose new positioning relative to old bounding box. This property is similar to "align" property of native text fields except it has two dimensions.

Here is an example of using this alignment property. The "original text" gets edited, scale and content are changed. Now user can affect the resulting positioning scheme. 

New GTK plugin window:
g7354-10
Positioning after edition: 
g7354-11

I modified only GTK version of plugin window, but TK still works with limited functionality, namely 
- no alignment property, default value ("middle center") always used
- scale step is too big, so edited object may not exactly match in size original

Besides that its fully functional.

Note: this patch includes some cleanup, replacing hand-written procedures with analogues from inkscape.

Also this patch should not break documents which was build with older version of this plugin (I did some basic compatibility tests, but may be more extensive testing is required). 

Please, let me know if there are any issues with this PR.

### DIFF
--- a/asktext.py
+++ b/asktext.py
@@ -488,7 +488,7 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
             self.global_scale_factor = self._scale_adj.get_value()
 
             try:
-                self.callback(self.text, self.preamble_file, self.global_scale_factor)
+                self.callback(self.text, self.preamble_file, self.global_scale_factor,self._alignment_combobox.get_active_text())
             except StandardError, error:
                 import traceback
 
@@ -629,10 +629,28 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
             scale_box = gtk.HBox(homogeneous=False, spacing=2)
             scale_frame.add(scale_box)
             self._scale_adj = gtk.Adjustment(lower=0.1, upper=10, step_incr=0.1, page_incr=1)
-            self._scale = gtk.HScale(self._scale_adj)
+            self._scale = gtk.SpinButton(self._scale_adj)
             self._scale.set_digits(1)
             self._scale_adj.set_value(self.scale_factor_after_loading())
             self._scale.set_tooltip_text("Change the scale of the LaTeX output")
+
+            liststore = gtk.ListStore(str)
+            for a in [  "top left","top center","top right",
+                        "middle left","middle center","middle right",
+                        "bottom left","bottom center", "bottom right"]:
+                liststore.append([a])
+
+
+            self._alignment_combobox = gtk.ComboBox()
+
+            cell = gtk.CellRendererText()
+            self._alignment_combobox.pack_start(cell)
+            self._alignment_combobox.add_attribute(cell, 'text', 0)
+            self._alignment_combobox.set_model(liststore)
+            self._alignment_combobox.set_wrap_width(3)
+            self._alignment_combobox.set_active(4)
+
+
 
             # We need buttons with custom labels and stock icons, so we make some
             reset_scale = self.current_scale_factor if self.current_scale_factor else self.global_scale_factor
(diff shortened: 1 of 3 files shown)


jcwinkler:
Thank you very much for this huge bunch of work and the detailed description, :+1: 

I will perform the necessary tests in my environment as soon as possible and will give you feedback.

Regards, Jan

P.S. We have really overseen `simpletransform.py` all the years... Would have saved a lot of work... :weary:

leberwurstsaft:
Regarding `simpletransform.py`, that's something I had noticed a while back but obviously forgotten right away :sleeping:.

Regarding your contribution: Wow, that looks great indeed :+1:

jcwinkler:
I tested the patch under Inkscape 0.92.2, 0.91.1 and 0.48.5 (Windows, Linux) while creating new nodes as well as while editing nodes created in previous versions of TexText. I did not observe any issues related to this patch. Everything works flawlessly and as expected.  

The patch also removes the little jumping of nodes after recompiling them (resulted from a small mistake in the handwritten calculation of the bounding box, I already fixed that on [https://bitbucket.org/jcwinkler/textext/branch/feature/pdf2svg_support](https://bitbucket.org/jcwinkler/textext/branch/feature/pdf2svg_support), but `simpletransform.py` is the cleaner way, of course. 

For the user interface I suggest to give the anchor position box a separate label so the user knows immideately what is meant. Maybe by puttin it into s separate group box. The TkInterface needs proper adaption, too, of course. But this can be done before the final release.

Hence, I suggest to merge this patch into the master.

I attached the 0.8 milestone so we can release this together with the pdf2svg, xetex and luatex stuff which will be most likely be completed next week. 

Is there any special way you would like to be credited, @sizmailov ? Or would you like to see your GitHub pseudonym only? Thank you very much again for your contribution!

sizmailov:
Sure, anchor position should go to a separate section, I used an ad-hoc solution so it's far from perfect.

You could mention me as "Sergei Izmailov"

I have a couple of GUI improvement suggestions. I'm not sure whether it's better to create new issue(s) or discuss it here, so I'll just mention them here:
 1. Don't show or disable "Reset" button for newly created nodes, since it does nothing except generating non-sense error messages when Ok button gets pressed (at least in GTK)
 2. Also "Global" label is misleading, since it actually sets scale to last used. Not sure how to name such action, "Set as previous" sounds right, but it's a bit wordy. Again, don't show button if there is no "previous" node. 
 3. Now error messages from preview show up again when "Ok" button pressed, even if all errors were fixed (at least in GTK). It should be easy to fix.

leberwurstsaft:
„At least in GTK“ sounds like the error messages are missing when using the TK variant. We should investigate that, too.

sizmailov:
I'm sorry, if was not clear about error messages.  

All error messages I'm talking about are either duplicates (p. 3) and should not appear or should not happen (p. 1) at first place.  "At least in GTK" means that I saw this erroneous error messages in GTK, but I did't try Tk. Presumably problems are common for GTK and Tk. 

Maybe I'll manage to fill the issue and write patch to it in the next couple of days.

jcwinkler:
I opened separate issues for the the GUI and error message topics since they are not related with this pull request.

jcwinkler:
Sergei, please wait with any new patches until I have merged the next bunch of code relating pdf2svg. This will introduce major changes and cleanup. Thank you!

Issue #15, opened 2018-04-06
Recompiling of nodes created with TexText < 0.5 fails


jcwinkler:
When editing nodes created with TexText < 0.5 the following error is thrown:

```
TypeError: unsupported operand type(s) for *=: 'NoneType' and 'float' 
```

Reason: These nodes do not have the scale attribute. Hence, the following code fails, since `current_scale` is `None` (returned by `TexText.get_old()`):

```
current_scale *= self.uutounit(1, "pt")
```

Possible fix: Let `TexText.get_old()` return 1.0 instead of None for the default scale factor.

jcwinkler:
**Hotfix for TexText 0.7.2**

Download link for fixed textext.py to replace installed file from version 0.7.2: [textext.py](https://github.com/textext/textext/blob/releases/0.7/textext.py)

jcwinkler:
Fixed in release 0.7.3

Issue #2, opened 2018-01-19
Migration to pdf2svg


jcwinkler:
TexText uses pstoedit for the conversion of the LaTeX generated pdf snippet to a svg node. Pstoedit itself depends on calls to ghostscript. Unfortunately, a certain option has been marked deprecated in ghostscript > 9.20 which is used by pstoedit. Hence, generation of the pdf snippet fails.

As an alternative we have decided to give the possibility to use another converter: pdf2svg. Work is currently in progress.

See [Issue 48](https://bitbucket.org/pitgarbe/textext/issues/48/ghostscript-still-bug-under-linux) and [Issue 57](https://bitbucket.org/pitgarbe/textext/issues/57/pdf2svg-migration).

jcwinkler:
Current status of development can be found at branch [feature/pdf2svg_support](https://github.com/jcwinkler/textext/tree/feature/pdf2svg_support) at my fork until pull-request #20 goes into this repo. Requires that pdf2svg is in your path. Tk interface not working 100%.

sizmailov:
Is `pstoedit` still supported as a backup converter or it completely deprecated?

sizmailov:
Right now plugin requires `pdf2svg` and does not fallback to `pstoedit` if `pdf2svg` is not found.

jcwinkler:
This is correct, as stated in pull request #20:

> Code works only if pdf2svg is found on the path. To use pstoedit comment/uncomment CONVERTERS in lines 1199/ 1200. (Will be in one list in final release).

sizmailov:
Indeed, I'm sorry, overlook that.

jcwinkler:
Provided in release 0.8.

Issue #3, opened 2018-01-19
Provide XeTeX and LuaTeX support


jcwinkler:
XeTeX and LuaTeX become more and more prominent, especially due to their native UTF-8 support. It would be nice if TexText supports snippets for these compilers.

jcwinkler:
The current development of this feature can be tracked in the branch [feature/xetex_luatex_support](https://github.com/jcwinkler/textext/tree/feature/xetex_luatex_support). It will be merged back as far as it is production ready to use.

leberwurstsaft:
Was the TK interface done in the merged pull request?  The description initially said, it wasn't.
In case it is still not done, we should leave this open. Or close this, but add another issue specific to that missing part.

jcwinkler:
At this moment, I am working on the TK implementation for xetex/luatex and anchor alignment.

jcwinkler:
Provided as of release 0.8. (Why did automatic close did not work? :angry: )

leberwurstsaft:
@jcwinkler You can automatically close issues with commit messages like this: https://help.github.com/articles/closing-issues-using-keywords/

Or with pull request descriptions like this: https://blog.github.com/2013-05-14-closing-issues-via-pull-requests/

jcwinkler:
I thought I did so, but I consequently wrote e.g. `resolves issue #3` (as in this case) instead of `resolves #3` as I am aware now after reading the links you provided :sob:

PR #20, opened 2018-05-08
Feature/pdf2svg support

jcwinkler:
This pull-request introduces the support of pdf2svg for converting the pdf output to svg. It resolves issue #2 and is a SUMO request. I know this is not what one really likes to review. However, major changes needed to be introduced into the code in order to use both, pstoedit as well as pdf2svg, as converters and defining an abstract interface for utilizing additional converters without tons of ifs and elses.

**Short summary:**
- New class `Pdf2SvgPlotSvg` for conversion pdf -> svg
- All methods in class `TexText` that are acting on the svg node moved to new class `SvgElement` encapsulating the svg node coming from `svg_to_group` and providing manipulation stuff
- `svg_to_group` of `pdf2svgPlotSvg`/ `PstoeditPlotSvg` returns an instance of `SvgElement` instead of xml node
- 2 subclasses of `SvgElement` providing pdf2svg and pstoedit specific stuff so we do not need to care about where the nodes coming from in the `TexText` class (especially useful when changing nodes via pdf2svg which have been originally created by pstoedit)
- Scaling feature from @sizmailov already included
- This pr also resolves issue #15 (compiling very old nodes, needed this patch for proper testing)

**Open:**
- Preserve colors defined in Inkscape
- Tk interface

Code works only if pdf2svg is found on the path. To use pstoedit comment/uncomment CONVERTERS in lines 1199/ 1200. (Will be in one list in final release).

**Long:**
Up to now, the workflow in `TexText` was as follows: The `PstoeditPlotSvg` class (derived from `LatexConverterBase`) returned an xml node via the method `svg_to_group`. The TexText class provided several methods for transforming, scaling and translating this node to ensure that it is placed correctly in the document.

Since svg code produced by pdf2svg fundamentally differs from svg code produced by pstoedit a new class `Pdf2svgPlotSvg` (derived from `PdfConverterBase`) has been introduced which does the conversion. Furthermore, since handling of the nodes must be partially specific to the used converter the `TexText` class would need to provide specific methods, too. This would blow the class up.

Hence, I introduced a new class called `SVGElement`, which encapsulates the xml node and provides all methods required for handling them. Converter specific methods are implemented in the subclasses `Pdf2SvgSvgElement` and `PstoEditSvgElement`. Then, the TextText class is only acting on instances of these classes and does not need to know where they come from. This is especially useful when modifying nodes using `pdf2svg`, which have been originally created by `pstoedit`.

My next step is to implement the possibility to keep color defined in Inkscape instead of using the LaTeX color. Furthermore, XeLaTex and LuaLaTex (issue #3) support will be added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,12 @@
 *.tar.gz
 tests/out.svg
 *.iss
+bezmisc.py
+cubicsuperpath.py
+inkex.py
+simplepath.py
+simplestyle.py
+simpletransform.py
 /PyGTK_for_Inkscape/0.92.2/build
 /PyGTK_for_Inkscape/0.92.2/download-*/
 /PyGTK_for_Inkscape/0.92.2/files/
--- a/textext.py
+++ b/textext.py
@@ -220,7 +220,6 @@ class TexText(inkex.Effect):
         if old_node is not None and old_node.is_textext_attrib("alignment"):
             alignment = old_node.get_textext_attrib("alignment")
 
-
         # Ask for TeX code
         if self.options.text is None:
             global_scale_factor = self.options.scale_factor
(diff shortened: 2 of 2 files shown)


leberwurstsaft:
Sorry, I have little time at the moment, between a baby girl and work, I can't always comment.
That said, this looks good. The decision to keep the nitty gritty out of the TexText class is correct. When the need arises to consider even better architecture, it should be simple enough given the right first steps.

jcwinkler:
This is 100% OK, of course. All the best!

Issue #16, opened 2018-04-07
Suggestions for GUI improvements


jcwinkler:
(Originally posted by @sizmailov in pull request #14):
 
1. Don't show or disable "Reset" button for newly created nodes, since it does nothing except generating non-sense error messages when Ok button gets pressed (at least in GTK)
 2. Also "Global" label is misleading, since it actually sets scale to last used. Not sure how to name such action, "Set as previous" sounds right, but it's a bit wordy. Again, don't show button if there is no "previous" node.

jcwinkler:
![grafik](https://user-images.githubusercontent.com/11866252/39759364-c3e37948-52d2-11e8-8eb1-5ee8352a981a.png)

Changes
- Scaling and Alignment in two different boxes
- Selection of TeX converter

What remains open:
- Naming of "Reset" and "Global":
   - `Reset` actually means "Reset to value used for creating this node"
   - `Global` actually means "Use scale factor of previously edited node"
   - Any short word for describing this? 
   - The tooltips do it but they are not available in the TK version and, personally, I am not willing to implement them via additional classes like Pmw ...
- Alternative title of groupbox "Alignment": "Alignment to old node" or something like this?

Any comments?

sizmailov:
What are actual use cases for `Global` button? 

I can imaging two potential scenarios:
1) Create several nodes in a row with a given scale factor
    Now scale factor gets initialized with previous value, so what button `Global` actually does is *resets* value to original. Let's remove `Global` button for fresh nodes and let `Reset` button do reset. 
2) Adjust scale of current node to match previous node
    This allows one to avoid copy-paste scale factors across nodes. Actually it's scale factors history of length 1. We can store several (say, up to 5) previously used values and allow user to pick one of them, see snapshot below. 

**Label names**
Now low precision in button labels is misleading, I suggest to remove it, i.e. "Reset (1.0)" would be just "Reset". 

- `Reset` actually means "Reset to value which was when window appeared", so just "Reset" is fine. 
- `Global` "As previous" lacks verb, but intent is still clear

The final window may look like this:
![image](https://user-images.githubusercontent.com/2975991/39785810-33fc84d8-5326-11e8-9fbd-086d1a1da293.png)

jcwinkler:
"As previous" is a very good suggestion! Thanks!

jcwinkler:
Yes, good idea!

PR #24, opened 2018-05-13
Tk interface: Added TeX command selection and alignment control

jcwinkler:
This pull request offers an update of the Tk interface so it matches the functionality of the GTK interface. I know it is not a candidate for a design award. But it does what it is meant for.

grafik

Since combo boxes are not available in the Tk version shipped with Inkscape on Windows / Python 2.7 I used radio buttons.

Furthermore, I renamed the label of the `global` scale factor button to `From previous`according to the discussion in issue #16.

The scale buttons do show the full resolution of the factor now. (Do we really need 3 decimals?)

Open: Disable alignment controls when creating a new node (Tk and GTK).

### DIFF
--- a/asktext.py
+++ b/asktext.py
@@ -738,8 +738,8 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
 
             # We need buttons with custom labels and stock icons, so we make some
             reset_scale = self.current_scale_factor if self.current_scale_factor else self.global_scale_factor
-            items = [('tt-reset', 'Reset ({0:.1f})'.format(reset_scale), 0, 0, None),
-                     ('tt-global', 'Global ({0:.1f})'.format(self.global_scale_factor), 0, 0, None)]
+            items = [('tt-reset', 'Reset ({0:.3f})'.format(reset_scale), 0, 0, None),
+                     ('tt-global', 'From previous ({0:.3f})'.format(self.global_scale_factor), 0, 0, None)]
 
             # Forcibly show icons
             settings = gtk.settings_get_default()
@@ -759,12 +759,12 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
 
             scale_reset_button = gtk.Button(stock='tt-reset')
             scale_reset_button.set_tooltip_text(
-                "Set scale factor to the value this node has been created with ({0:.1f})".format(reset_scale))
+                "Set scale factor to the value this node has been created with ({0:.3f})".format(reset_scale))
             scale_reset_button.connect('clicked', self.reset_scale_factor)
 
             scale_global_button = gtk.Button(stock='tt-global')
             scale_global_button.set_tooltip_text(
-                "Set scale factor to the value of the previously edited node in Inkscape ({0:.1f})".format(
+                "Set scale factor to the value of the previously edited node in Inkscape ({0:.3f})".format(
                     self.global_scale_factor))
             scale_global_button.connect('clicked', self.use_global_scale_factor)
 
--- a/asktext.py
+++ b/asktext.py
@@ -54,6 +54,7 @@ try:
 except ImportError:
     try:
         import Tkinter as Tk
+        import tkMessageBox as TkMsgBoxes
 
         TOOLKIT = TK
     except ImportError:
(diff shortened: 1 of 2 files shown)


sizmailov:
It does look nice)

> Do we really need 3 decimals?)

The precision mattes since we couple inkscape scales with textext. When we recreate node the scale factor from GUI is used so inaccuracy may become visible. 3 decimals places should be enough to make sure new node visually indistinguishably from old one. 

I agree, too many digits are distracting, but we need to keep functionality. 

If you don't resize your textext objects in inkscape 3 digits does not make any sense for you. In principle it's possible to find out suitable precision for particular object (say, 1 digits for non-scaled objects and 3 digits for inkscape-scaled objects).

To keep things simple and functional I would just keep 3 digits for all objects. 

> The scale buttons do show the full resolution of the factor now
No doubts, precision on scale buttons must match precision of spin box.

asktext.py
```diff
@@ -231,32 +258,60 @@ def ask(self, callback, preview_callback=None):
...
+            self._scale.delete(0, "end")
+            self._scale.insert(0, self.scale_factor_after_loading())
 
             reset_scale = self.current_scale_factor if self.current_scale_factor else self.global_scale_factor
-            self._reset_button = Tk.Button(box, text="Reset ({0:.1f})".format(reset_scale),
+            self._reset_button = Tk.Button(box, text="Reset ({0:.3f})".format(reset_scale),
                                            command=self.reset_scale_factor)
             self._reset_button.pack(ipadx=10, ipady=4, pady=5, padx=5, side="left")
-            self._global_button = Tk.Button(box, text="Global ({0:.1f})".format(self.global_scale_factor),
+            self._global_button = Tk.Button(box, text="From previous node ({0:.3f})".format(self.global_scale_factor),
```
sizmailov: I think "As previous" sounds better than "From previous"

sizmailov: Labels in GTK and TK versions must match

asktext.py
```diff
@@ -675,8 +738,8 @@ def create_window(self):
 
             # We need buttons with custom labels and stock icons, so we make some
             reset_scale = self.current_scale_factor if self.current_scale_factor else self.global_scale_factor
-            items = [('tt-reset', 'Reset ({0:.1f})'.format(reset_scale), 0, 0, None),
-                     ('tt-global', 'Global ({0:.1f})'.format(self.global_scale_factor), 0, 0, None)]
+            items = [('tt-reset', 'Reset ({0:.3f})'.format(reset_scale), 0, 0, None),
+                     ('tt-global', 'From previous ({0:.3f})'.format(self.global_scale_factor), 0, 0, None)]
```
sizmailov: Labels in GTK and TK versions must match

asktext.py
leberwurstsaft: Agreed

asktext.py
leberwurstsaft: 👍

leberwurstsaft:
> To keep things simple and functional I would just keep 3 digits for all objects.
Yes, don't add special cases. Consistency is key.

asktext.py
```diff
@@ -213,6 +217,29 @@ def __init__(self, text, preamble_file, global_scale_factor, current_scale_facto
...
+                invalid content. So with or without this validation method you need to delete the content of the
+                box before inserting the new stuff.
+            """
+            if S == '' or P == '':
+                # Initialization of widget (S=='') and deletion of entry (P=='')
+                valid = True
+            else:
+                # All other cases: Ensure that result is OK
+                try:
+                    float(P)
```
leberwurstsaft: interesting trick?

jcwinkler: S is the input coming from the user (a character from the keyboard or the content of the clipboard), s is the string currently in the box and P is the string which will be placed into the box after the input process ended. P is set by Tk.  

Content of box: s="1.23", User input: S=","  --> P = "1.23,"  --> Input rejected
Content of box s="1.23", User input: S="5.44" from clipboard --> P = "1.235.44" !! [Even without the validation method I had to delete the entry before entering a new one from clipboard. Selected entries are not overwritten as one would naturally expect. Strange Tk.]

Will document it.

jcwinkler:
Thanks for the comments. I will implement the suggestions as proposed.

asktext.py
jcwinkler: Done

jcwinkler:
I agree on the number of decimals. I hadn't in mind the issue coming from scaling in Inkscape since personally I do only scale via TexText. Thanks.

asktext.py
leberwurstsaft: I specifically meant the `float(P)` part to validate the input.

Issue #21, opened 2018-05-08
Improved color management


jcwinkler:
Currently the colors defined in the LaTeX code are used for displaying the node in Inkscape.

Some users like to assign the colors in Inkscape. When the node is recompiled this color assignment is lost. 

For the user it would be nice to have the possibility to decide if the color defined in Inkscape is preserved after recompiling or if the colors from the TeX code is used.

sizmailov:
Is it supposed to handle one color per textext object or its should handle colored elements too? 

Say I want to highlight sine argument `x`. What should I get after replacement `x` by `x+1`? What should be colored `x` or `x+1`?
![image](https://user-images.githubusercontent.com/2975991/39799696-d5f4652a-536d-11e8-9638-f602bcf9f28e.png)

I wonder how it might be implemented in general and reliable way? To do that you need to know one-to-one correspondence between children of old and new node, but this information is not available to textext.

jcwinkler:
What I am more worried about is the following behavior in pdf2svg nodes: After assigning the fill color to a node one may obtain the following:
![grafik](https://user-images.githubusercontent.com/11866252/39801594-02b2525e-536c-11e8-8f57-43a082da6555.png)

To get the root completely colored one needs to also set the stroke color as in pstoedit), but this leads to:

![grafik](https://user-images.githubusercontent.com/11866252/39801641-246ca494-536c-11e8-84b3-aa6ba1ff3b93.png)

Reason is the path desribing the bar of the root:
```
        <path
           id="path29"
           transform="matrix(1,0,0,-1,182.358,126.263)"
           d="M 0.001375 0.00128125 L 30.657625 0.00128125 "
           style="fill:none;stroke-width:0.398;stroke-linecap:butt;stroke-linejoin:miter;
                      stroke:rgb(0%,0%,0%);stroke-opacity:1;stroke-miterlimit:10;" />
```

which differs from the other elements which are of type

```
        <g
           id="g27"
           style="fill:rgb(0%,0%,0%);fill-opacity:1;">
          <path
             id="path25"
             transform="translate(174.056,126.462)"
             d="M 3.875 ... "
             style="stroke:none;" />
        </g>
```

Adding `stroke-width:0;` to each path-element which has `style="stroke:none;` resolves the problem. But I do not know if this is valid svg code.

jcwinkler:
(It would be easy to do this in the `svg_to_group` method of `Pdf2SvgPlotSvg` class.)

sizmailov:
[...]

PR #26, opened 2018-05-15
Keep Inkscape / TeX colors

jcwinkler:
This pull request relates to issue #21.

If svg code coming from the converter contains non-black styles -> TeX colorization is kept

If everything is black coming from the converter -> Inkscape top level group color/ opacity is preserved

See commit messages for more details.

Note that pstoedit colorizes its items using "fill" etc. while pdf2svg uses css styles (as Inkscape). Neverthelss I check for both ways in both converters `is_colorized` methods. Maybe too paranoid?

Tested both converters in Inkscape 0.92.

### DIFF
--- a/textext.py
+++ b/textext.py
@@ -956,7 +956,15 @@ class Pdf2SvgPlotSvg(PdfConverterBase):
             for node in svg_raw:
                 if node.tag != "{%s}defs" % SVG_NS:
                     new_group.append(node)
-            # return PsToEditSvgElement(copy.copy(tree.getroot().xpath('g')[0]))
+
+            # Ensure that strokes with color "none" have zero width to ensure proper colorization via Inkscape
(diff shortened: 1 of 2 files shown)


sizmailov:
Good job. Tested on ubuntu with ps2edit.

textext.py
```diff
@@ -394,6 +392,11 @@ def do_convert(self, text, preamble_file, user_scale_factor, converter_class, ol
             relative_scale = user_scale_factor / original_scale
             new_svg_ele.align_to_node(old_svg_ele, alignment, relative_scale)
 
+            # If no non-black color has been explicitely set by TeX we copy the color information from the old node
```
leberwurstsaft: just a typo. its "explicitly"

jcwinkler:
Thank you for the comments!

Issue #19, opened 2018-04-25
Missing width and height attributes in svg document lead to crash


Pierre-Andre:
- When using TeXtext on an empty document with $\beta_1$ TeX text it works (obviously).
- But when I use texText on a more complicated svg (created whith R and svglite device) the same textext object throw the following error:
```
--------------------------------------------------
expected string or buffer
Traceback (most recent call last):
  File "/home/pac/.config/inkscape/extensions/asktext.py", line 491, in cb_ok
    self.callback(self.text, self.preamble_file, self.global_scale_factor)
  File "textext.py", line 224, in <lambda>
    old_node),
  File "textext.py", line 351, in do_convert
    width = self.unittouu(root.get('width'))
  File "/usr/share/inkscape/extensions/inkex.py", line 387, in unittouu
    p = param.match(string)
TypeError: expected string or buffer
----------------------------------------------------
```
I do suspect a problem with getting the size of the page (and may be a default/fallback width can be useful ?)

leberwurstsaft:
Good point. Width and height are not required values and are implicitly assumed to be "100%" when missing. Question is, what would 100% refer to when it is the rood node?

The recommended way to specify dimensions according to some article I read is to define a viewBox anyway. So we could also look at the viewBox, either as a fallback or as the primary way to retrieve the doc's dimensions. It looks like in the example by @sizmailov the viewBox uses document units rather than mm or anything from the real word?

Pierre-Andre:
The svg is not created by inkscape and it does not provide height and width. It does provide a viewBox (in pixels, even it is not explicitely written in svg files, inkscape understand it as pixels):

```<?xml version='1.0' encoding='UTF-8' ?>
<svg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' viewBox='0 0 198.43 198.43'>
<defs>
  <style type='text/css'><![CDATA[
    line, polyline, polygon, path, rect, circle {
      fill: none;
      stroke: #000000;
      stroke-linecap: round;
      stroke-linejoin: round;
      stroke-miterlimit: 10.00;
    }
  ]]></style>
</defs>
<rect width='100%' height='100%' style='stroke: none; fill: #FFFFFF;'/>
<defs>
'''
Of course, when I modify the document properties (ctrl+maj-d) I can resize page dimension to content (which insert height and width in svg) and TeXtext will work as desired, but I do think that it should work with viewbox only...

jcwinkler:
[...]

PR #27, opened 2018-05-16
Handle documents without width/height and/or viewbox specification

jcwinkler:
Resolves issue #19

Document size determination during TexText node creation is now being
done using `inkex.Effect.getDocumentHeight()` and
`inkex.Effect.getDocumentWidth()` methods which provide the necessary
fallback stuff if width/height or viewbox attributes are missing.

Wrapped methods into `TexText.get_document_height()' and
'TexText.get_document_width()` methods to provide compatibility with
Inkscape 0.48. Same for `uutounit` and `unittouu` to avoid repeated
try-except blocks in the processing code.

### DIFF
--- a/textext.py
+++ b/textext.py
@@ -174,6 +174,50 @@ class TexText(inkex.Effect):
             dest="scale_factor",
             default=self.settings.get('scale', float, 1.0))
 
+    # Identical to inkex.Effect.getDocumentWidth() in Inkscape >= 0.91, but to provide compatibility with
+    # Inkscape 0.48 we implement it here explicitly again as long as we provide compatibility with that version
+    def get_document_width(self):
+        width = self.document.getroot().get('width')
+        if width:
+            return width
+        else:
+            viewbox = self.document.getroot().get('viewBox')
+            if viewbox:
+                return viewbox.split()[2]
+            else:
+                return '0'
+
+    # Identical to inkex.Effect.getDocumentHeight() in Inkscape >= 0.91, but to provide compatibility with
+    # Inkscape 0.48 we implement it here explicitly again as long as we provide compatibility with that version
+    def get_document_height(self):
+        height = self.document.getroot().get('height')
+        if height:
+            return height
+        else:
+            viewbox = self.document.getroot().get('viewBox')
+            if viewbox:
+                return viewbox.split()[3]
+            else:
+                return '0'
+
+    def unit_to_uu(self, unit):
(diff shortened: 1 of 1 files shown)


jcwinkler:
I tested this with documents having no width/height spec and documents which even do not have a viewbox defined and it seems to work as exepcted.

textext.py
```diff
@@ -174,6 +174,50 @@ def __init__(self):
...
+    def get_document_width(self):
+        width = self.document.getroot().get('width')
+        if width:
+            return width
+        else:
+            viewbox = self.document.getroot().get('viewBox')
+            if viewbox:
+                return viewbox.split()[2]
+            else:
+                return '0'
```
sizmailov: I would go with try-catch approach, instead. You may not account some edge case with `if` statements, so `try-catch` is safer here.

textext.py
```diff
@@ -174,6 +174,50 @@ def __init__(self):
...
+    def get_document_height(self):
+        height = self.document.getroot().get('height')
+        if height:
+            return height
+        else:
+            viewbox = self.document.getroot().get('viewBox')
+            if viewbox:
+                return viewbox.split()[3]
+            else:
+                return '0'
```
sizmailov: ditto

textext.py
leberwurstsaft: In general, I'd agree, but I think, since this is verbatim the same code as in `inkex.py` itself, we can "safely" assume that when it has not yet caused issues for anyone, then it won't for us.

To be safe, go try/catch. Your call.

textext.py
```diff
@@ -174,6 +174,50 @@ def __init__(self):
...
+
+    def unit_to_uu(self, unit):
```
leberwurstsaft: Good idea to wrap these.

textext.py
```diff
@@ -308,12 +347,7 @@ def do_convert(self, text, preamble_file, user_scale_factor, converter_class, ol
 
         # Coordinates in node from converter are always in pt, we have to scale them such that the node size is correct
         # even if the document user units are not in pt
-        try:
-            # Inkscape > 0.48
-            scale_factor = user_scale_factor*self.unittouu("1pt")
-        except AttributeError:
-            # Inkscape <= 0.48
-            scale_factor = user_scale_factor*inkex.unittouu("1pt")
+        scale_factor = user_scale_factor*self.unit_to_uu("1pt")
```
leberwurstsaft: I prefer operators to have single spaces around them. The `*` misses them.

PR #41, opened 2018-08-23
Fix for user preamble file reset, added file chooser button in Tk

### LINKED ISSUE #40: Non-default preamle is reset when text dialog is opened
When a non-default preamble is selected in the text dialog, it works for compiling and inserting into the document. It is also properly saved in the SVG file. However, when the dialog is opened again to edit the text, the preamble is reset to the default one (default_packages.tex). The reason is that clear_preamble is called during the window creation after the old preamble value is set in the constructor. The attached patch fixes the problem.
[asktext.py.patch.txt](https://github.com/textext/textext/files/2310331/asktext.py.patch.txt)

jcwinkler:
Fixes #40


leberwurstsaft:
Good catch. 

Btw:  Once, I was thinking about storing the whole content of the preamble that was used inside the SVG, so that it can even be found when editing the document on another computer. But there are all kinds of trouble ahead when trying this and the needed packages are not on the other computer etc. So, not really an option.

PR #53, opened 2018-10-24
Cleanup repository

### LINKED ISSUE #28: Improve installation script (setup.py)
In release pack there is `./extension` directory while it's absent git repository. 

Right now `setup.py` copies all files from current directory to inkscape extension directory. 
It used to copy all from `./extension` folder before #14 (I "fixed" that, since I was not aware of release distribution structure). 

This should be fixed, so script would work for both dev and release distributions.

sizmailov:
Repository is restructured. 

All extension files are moved to `./extension`. 
All extension python scripts are moved to in `./extension/textext` in order to keep inkscape extension folder clean and avoid potential clashes with other extensions. 

`setup.py` is rewritten and works from repository and distribution. 
`build_linux_package` is now a python script, and could be used from any OS. 

The extension code itself is untouched (`textext.py` renamed to `__init__.py` and paths are tweaked in `textext.inx`)

Close #28


sizmailov:
@leberwurstsaft, @jcwinkler Could you please review changes? 

I'm unaware of purpose/usage of `./tests`. Is there any automatic unit tests? 
Probably now they are broken, so fix needed before merge. 

`installer.nsi` seems to be auto-generated, so it needs update too.

jcwinkler:
I created a new branch new-repo-structure where the new stuff can be found for testing.

jcwinkler:
> I'm unaware of purpose/usage of `./tests`. Is there any automatic unit tests?
> Probably now they are broken, so fix needed before merge.

Except for the win_app_paths tests the files are not needed anymore (the were used for testing of the manual calculation of the bounding box). I will remove them.

jcwinkler:
> `installer.nsi` seems to be auto-generated, so it needs update too.

In fact, it is hand coded based on an autogenerated version. I will adapt the installer as far as I have evaluated the new structure in Windows.

Issue #17, opened 2018-04-07
Error messages from preview reappear after compilation


jcwinkler:
(Originally posted by @sizmailov in pull request #14 but not related specifically to this pull request)

When LaTeX error messages occur during writing code and checking it using the preview feature they reappear when the code has been finally and successfully compiled.

This is due to the way the error handling queue is processed inTexText.

PR #62, opened 2018-11-06
Improve error handling + setup.py update

sizmailov:
## Error handling 
 + Added a logfile
 + Top-level error collection 
 + Exceptions are in hierarchy and more telling
 + Don't show same error twice (#17)
 + Suppress additional warning window in kde (missing theme)
 + Reduce `return None -> check for None -> raise` to simple `raise` at first place.

todo: if fatal error is unavoidable show it right away

## setup.py
  + each check marked with Succ/Fail/Ukwn (helpful if terminal does not support colors)
  + add option to suppress console colors
  + remove old installation (if any) by default
  + ask for preamble replacement confirmation iff old file exists and differs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,18 +24,6 @@ matrix:
           - texlive-latex-base
 
     - os: linux
-      dist: precise
-      addons:
-        apt:
-          packages:
-          - pstoedit
-          - pdf2svg
-          - inkscape
-          - python-gtk2-dev
-          - imagemagick
-          - texlive-latex-base
-
-    - os: linux
       dist: xenial
       addons:
         apt:
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -58,6 +58,7 @@ import StringIO
 import subprocess
 import sys
 import tempfile
+import traceback
 
 
 class ChangeDirectory(object):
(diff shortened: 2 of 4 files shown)


sizmailov:
Github shows changes with default diff algorithm.  `__init__.py` appears to be completely rewritten while it is not the case. `git difftool --minimal <rev1> <rev2>` produces sensible diff.

leberwurstsaft:
> Github shows changes with default diff algorithm. `__init__.py` appears to be completely rewritten while it is not the case. `git difftool --minimal <rev1> <rev2>` produces sensible diff.

I was wondering the same thing. It looks like it was completely altered, which of course isn't the case. But there are whitespace changes all over the place which, I believe, is the reason.

sizmailov:
@leberwurstsaft It's purely algorithm problem, github has option to ignore whitespace changes:

image

jcwinkler:
Sorry guys, sth. is really messed up here. @sizmailov please reopen this pull-request in such a way that it can be merged into the develop branch. Maybe you have to adjust some things on your side. Then I will see how to continue. Too many things in parallel here. My fault. Thanks!

sizmailov:
ok, no problem

jcwinkler:
Thanks, Sergei!

PR #68, opened 2018-11-10
Fix issue 35

sizmailov:
+ Add `previous_error_code` global setting to avoid massive logging on well configured setups.
+ Move part of `setup.py` to `textext/requirements_check.py`
+ Do requirements check before any user input (apply same set of rules as in `setup.py`).
Skip this check if previous run was successful. 
+ Produce readable nested log structure with context managers. Automatic trailing `done`/`failed` is added on context exiting. Example:
```txt
[2018-11-10 04:06:08,601][DEBUG   ]:       Converting tex to svg          //  __init__.py:625  
[2018-11-10 04:06:08,601][DEBUG   ]:         Converting .tex to svg element          //  __init__.py:1011 
[2018-11-10 04:06:08,633][DEBUG   ]:           Converting .tex to .pdf          //  __init__.py:947  
[2018-11-10 04:06:08,787][DEBUG   ]:           Converting .tex to .pdf done          //  __init__.py:977  
[2018-11-10 04:06:08,787][DEBUG   ]:           Converting .pdf to .svg          //  __init__.py:1136 
[2018-11-10 04:06:08,797][DEBUG   ]:           Converting .pdf to .svg done          //  __init__.py:1144 
[2018-11-10 04:06:08,797][DEBUG   ]:           Groupping resulting svg          //  __init__.py:1159 
[2018-11-10 04:06:08,798][DEBUG   ]:           Groupping resulting svg done          //  __init__.py:1206 
[2018-11-10 04:06:08,799][DEBUG   ]:         Converting .tex to svg element done          //  __init__.py:1021 
[2018-11-10 04:06:08,799][DEBUG   ]:       Converting tex to svg done          //  __init__.py:626  
[2018-11-10 04:06:08,799][DEBUG   ]:       Replacing node in document          //  __init__.py:656  
[2018-11-10 04:06:08,805][DEBUG   ]:       Replacing node in document done          //  __init__.py:665
```


sizmailov:
Heavy logging would be a performance hit as well as unnecessary requirements checks. We can't do it all the time. 

The worst scenario is our baseline. GUI (PyGTK or Tk) may not even appear in some circumstances, so we are limited to interface provided by inkscape. 

So logging level should be adapted as needed. We can't do it with GUI switch. PyGtk/Tk may not be even installed. The `.inx` parameters is not the way also. 

To address this problem new global parameter `previous_error_code` added. So log level is adjusted at initialization based on previous run. Requirements are conditionally checked based on this parameter. 

I didn't log explicitly any information about version of inkscape/pstoedit/latex, etc since it would appear naturally in requirements check.

Right now logged information seems to be sufficient to realize the source of most of previously reported issues. Please tell if there is anything missing.

jcwinkler:
Good catch! Especially the `previous_exit_code` idea! I am evaluating it on my systems and will give you feedback as soon as possible.

PR #71, opened 2018-11-20
Improved setup procedure/ Add lookup handles

jcwinkler:
If this is OK for you @sizmailov I would like to merge this into the repo.

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -347,9 +347,7 @@ try:
                                 original_scale=current_scale
                                 )
 
-
-        @staticmethod
-        def preview_convert(text, preamble_file, converter_class, image_setter, tex_command):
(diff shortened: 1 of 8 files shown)


extension/textext/__init__.py
```diff
@@ -1532,19 +1159,19 @@ def _get_flip_transformation(self, ref_node):
     logger.info("If problem persists, please file a bug https://github.com/textext/textext/issues/new")
     user_log_channel.show_messages()
     try:
-        settings = Settings()
-        settings.set("previous_exit_code", EXIT_CODE_UNEXPECTED_ERROR)
-        settings.save()
+        cache = Cache()
```
sizmailov: Store `previous_exit_code` in cache, instead of config

Issue #66, opened 2018-11-08
Possibility to run setup.py on Windows


jcwinkler:
This issue is related to #28. Goal is to drive `setup.py` on Windows even if no standalone Python distribution is installed on the system. This can be achieved by using the Python distribution which is shipped with Inkscape. For this purpose setup.py is called from a batch file.

- [x] Reliable determination of the Inkscape installation for calling Python
- [x] Handle colors of setup.py output (at least disable them)
- [x] Adjust `win_app_paths.py`
- [x] Make sure executable tests run under Windows
     - [x] ghostscript
     - [x] pstoedit
     - [x] pdf2svg
     - [x] latex, xelatex, lualatex
     - [x] python (+TkInter, PyGTK, ...)
- [x] Passing of command line arguments

PR #73, opened 2018-11-23
Feature/update win app paths

jcwinkler:
By this PR full compatibility of `setup.py` on Windows systems is introduced, see issue #66  (driving setup.py from a batch file will be a seperate PR).

The modifications done address the following properties of a Inkscape Windows installation:

1. It is not guaranteed that all executables we need are in the system path. Hence, `win_app_paths.py` offers a function which returns the necessary paths. By this the `WindowsDefault` class tweaks the os path so `TexTextRequirementsChecker` can proceed as under Linux, see commit 
348146b  ( `win_app_paths.py` has been completely rewritten)

2. Under Windows there exist possibly several executable names for the same program. Candidate is Ghostscript which can be named as `gswin64c.exe` or `gswin32c.exe` or `gs.exe`. To account for this the executable name can also be provided as a list of strings in the `Default` classes. As a consequence, the `find_executable` and `_find_executable_in_path` methods have to be able to iterate over several executable names of the same program. See commit 4773f02 

3. Under Windows 10 colored output is enabled if the anniversary update is detected, see commit 
a43d97f  (Colorama is no option since it is not in Inkscape's Python distro.

4. The ´subprocess.check_call` function always opens a command window when run from Inkscape. This can be supressed if the `Popen` / `communicate` approach is used instead, see commit 
eda1d44 

Open is the detection of Python in older versions of Inkscape (there it is not in the Inkscape base dir).

I tested the stuff under Windows 7, Windows 8.1 and Windows 10.

I will offer the batch file for driving this script (https://github.com/jcwinkler/textext/blob/96736248edca2e21d2acf7551d4abc2f92d19ced/setup_win.bat) in a separate PR, needs some final testing.


jcwinkler:
grafik

jcwinkler:
(OK, I have identified the source of the failed TRAVIS check. I will fix this later in the day...)

extension/textext/requirements_check.py
```diff
@@ -12,27 +12,102 @@ class Defaults(object):
...
+            # -> https://docs.microsoft.com/en-us/windows/console/getstdhandle
+            h_stdout = h_kernel32.GetStdHandle(-11)
+
+            # ENABLE_PROCESSED_OUTPUT  | ENABLE_WRAP_AT_EOL_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING = 7
+            # -> https://docs.microsoft.com/en-us/windows/console/setconsolemode
+            result = h_kernel32.SetConsoleMode(h_stdout, 7)
+
+            console_colors = "always"
+        except (ImportError, AttributeError):
+            pass
```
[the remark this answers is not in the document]
sizmailov: Move all executable code into `__init__`, so it would be executed on WIndows platform only

extension/textext/requirements_check.py
```diff
@@ -519,34 +590,45 @@ def find_pstoedit(self, version=None):
...
+    @staticmethod
+    def _check_call(command):
+        # Ensure that command window does not pop up on Windows!
+        info = subprocess.STARTUPINFO()
+        info.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+        info.wShowWindow = subprocess.SW_HIDE
+        p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=info)
+        return p.communicate()
+
+    def _find_executable_in_path(self, executable_names):
```
sizmailov: rename argument to `program_name`

extension/textext/requirements_check.py
```diff
@@ -519,34 +590,45 @@ def find_pstoedit(self, version=None):
...
         messages = []
```
sizmailov: fetch `executable_names` from `defaults`

extension/textext/requirements_check.py
```diff
@@ -659,6 +741,7 @@ def help_message_with_url(section_name, executable_name=None):
 get_levels_colors = LoggingColors()
 
 if sys.platform.startswith("win"):
+    import win_app_paths as wap
```
sizmailov: move import into `WindowsDeafults.__init__()`

extension/textext/requirements_check.py
```diff
@@ -519,34 +590,45 @@ def find_pstoedit(self, version=None):
                     "pstoedit=%s is not found (but pstoedit=%s is found)" % (version, found_version)])
         return RequirementCheckResult(None, ["Can't determinate pstoedit version"])
 
-    def find_executable(self, executable_name):
```
sizmailov: revert changes to this function, all system-specific changes would be in `_find_executable_in_path`

extension/textext/requirements_check.py
```diff
@@ -432,14 +507,14 @@ def __init__(self, logger, config):
         self.available_tex_to_pdf_converters = {}
         self.available_pdf_to_svg_converters = {}
 
-        self.inkscape_executable_name = "inkscape"
```
sizmailov: We need to have both names of programs and names of actual executables. Let's define program names in TexTextRequirementsChecker, and executable names in Windows/Linux/MacOsDefaults

Names of programs MUST match names in command line arguments. `"ghostscript" -> "--ghostscript-executable=C://.../gswin64.exe"`

First. Need to rename `*_executable_name` in TexTextRequirementsChecker to `*_name` and let them be strings as before.

Second. Let's define dictionary in Defaults which would map name to executable name list. 

linux mapping would be 

```python
class LinuxDefaults(Defaults):
    os_name = "linux"
    console_colors = "always"
    executable_basenames = {
        "inkscape": ["inkscape"],
        "python2.7": ["python2.7"],
        "pdflatex": ["pdflatex"],
        "lualatex": ["lualatex"],
        "xelatex": ["xelatex"],
        "pdf2svg": ["pdf2svg"],
        "pstoedit": ["pstoedit"],
        "ghostscript": ["ghostscript"]
    }

```

jcwinkler:
Thanks for the suggestions. I will go through them and update the PR.

jcwinkler:
I implemented the distinction between program and executable names so that command line options work now.

Honestly speaking, I am not 100% happy with this approach since initialization of the properties is executed for both, `WindowsDefaults` and `LinuxDefaults`, always on any platform although the class is not instantiated. I will provide a slightly modified approach in a separate branch for evaluation. (This is all a matter of taste, I know. The current implementation works and that's good)

sizmailov:
If you want some code to be NOT evaluated on other platform - put it class methods, in `__init__`. 
Now most of LinuxDefaults properties are class properties, while `inkscape_extensions_path` is property of LinuxDefaults instance. 

We access only Default's instance properties which looks for class properties if instance property does not exist. 

Class properties are more readable I think, at least for LinuxDeafults.

Again, I suggest to move `try .. import win_app_paths ... catch` to WindowsDefaults and overwrite class property `console_colors` with instance property.

Note: you may want to mark review comments as resolved if you think they are.

sizmailov:
@jcwinkler Changes look good to me.

PR #75, opened 2018-11-27
Feature/batch file driven setup

### LINKED ISSUE #66: Possibility to run setup.py on Windows

jcwinkler:
This pull request offers the batch file which can be used on Windows to run setup.py using the Python interpreter shipped with inkscape. It resolves #66 .

In the most common case the user just has to double-click on `setup_win.bat`. The script tries to find Inkscape and its Python interpreter. Then it calls `setup.py` using the found interpreter.

If users have installed Inkscape from a zip package without adding the Inkscape location to the system path they can pass the path to Inkscape via a command line argument: 
```
setup_win /D:"C:\Path\to\Inkscape installation"
```

Furthermore it is also possible to pass the arguments understood by setup.py. e.g.
```
setup_win.bat /d:"C:\Program Files\Inkscape" /p:"--ghostscript-executable 'C:\Program Files\gs\gs.exe' --pstoedit-executable echo 'C:\Program Files\pstoedit\pstoedit.exe'"
```

I forgot what it means to write a Windows batch file which is to some extend "save" against all kind of user inputs. So don't expect elegant code. 

I also put several comments w.r.t. the crude batch file syntax so I remember this stuff when working on this file later.

grafik

### DIFF
--- a/setup_win.bat
+++ b/setup_win.bat
@@ -1,4 +1,64 @@
 @echo off
-rem ToDo: Proper detection of path to Inkscape
-rem       Passing of command line options
-"C:\Program Files\Inkscape\python" setup.py
+rem setup_win.bat
+rem Wrapper for setup.py on Windows systems using the Python installation
+rem shipped with Inkscape
+rem
+rem ToDo:
+rem -Pass command line args for setup.py
+rem -Allow to pass inkscape dir directly
+rem -Refer to different python location in Inkscape < 0.92.2
+
+setlocal EnableExtensions
+setlocal EnableDelayedExpansion
+set "INKSCAPE_DIR="
+set "INKSCAPE_CMD="
+
+rem Inkscape installation path is usually found in
+rem "Software\Microsoft\Windows\CurrentVersion\App Paths\inkscape.exe"
+rem under HKLM or HKCU
+for %%R in (HKLM HKCU) do (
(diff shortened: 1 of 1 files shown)


sizmailov:
Can we have same set of arguments for .bat and setup.py? it would be convenient.

jcwinkler:
I had this in mind but then abandoned it due to the stupid handling of whitespaces, :, = and -- as well as the work with dynamic structures in the batch file syntax (The command line arguments fall into atomic parts when whitespaces are found and we already need the double quotes for the path names wich may contain white spaces, too). 

Since the average user will only have to double-click on the file (or at most pass the D: option) while any serious developer will directly use the Python script I hope we can live with the current approach. When there is some more time I will try to improve it.

PR #76, opened 2018-11-28
Universal build package script and modified Windows installer

jcwinkler:


build_packages.py
```diff
@@ -0,0 +1,61 @@
...
+            shutil.rmtree(self.path)
+
+
+if __name__ == "__main__":
+
+    TexTextVersion = "0.8.1"
+
+    parser = argparse.ArgumentParser(description="Build TexText distribution archive for all supported platforms.")
+    available_formats = [fmt for fmt, desc in shutil.get_archive_formats()]
+    parser.add_argument('--linux',
```
sizmailov: Set of desired formats differs for different platforms. Linux should have tgz and zip, while for windows it's only zip. 
So currently script should be run twice, once for each platform to build only desired formats.

If you change options to indicate which formats should be build for each platform the script can be run once. Example: 
```bash
# build all at once
build_packages.py --linux-formats tgz zip --windows-formats zip 
# build windows only 
build_packages.py --windows-formats zip 
```

build_packages.py
```diff
@@ -0,0 +1,61 @@
...
+        args.platforms = ["Linux", "Windows"]
+
+    for platform in args.platforms:
+        PackageName = "TexText-%s-" % platform.capitalize() + TexTextVersion
+        git_ignore_patterns = shutil.ignore_patterns(*open(".gitignore").read().split("\n"))
+
+        with TmpDir() as tmpdir:
+            shutil.copytree("./extension",
+                            os.path.join(tmpdir, "extension"),
+                            ignore=git_ignore_patterns  # exclude .gitignore files
```
sizmailov: Should be win_app_paths.py excluded from non-windows distributions?

jcwinkler: Then we would have to provide a modified .inx file on Linux. We had this problem when we introduced `win_app_paths.py` and, hence, decided to ship `win_app_paths.py` also on Linux although it is never loaded there.

build_packages.py
jcwinkler: I would just leave the Windows tgz on the disk :smile: 

But I will improve the command line options in the suggested way.

build_packages.py
sizmailov: Ok, lets keep it.

build_packages.py
jcwinkler: I updated in that way. Not a "perfect" solution but I think it is enough.

PR #78, opened 2018-11-30
Use current python for requirement checks

sizmailov:
Note: adjustments of .bat script are required if merged.


sizmailov:
If python3 is selected in inkscape preferences the error message will complain about incompatible python version instead of cryptic import errors.
image

Issue #47, opened 2018-10-06
Wrap text in the latex text box


veltsov:
It would be nice to wrap (automatically move to a new line) the latex text in the editor box. I often copy the latex text from some other source, which might have very long lines due to automatic wrapping usually available in any editor. Such long lines are difficult to work with in the current textext dialog. Unconditionally the wrapping is added by a single-line patch (attached), but full implementation should probably include a checkbox in the menu.
[asktext.py.patch.txt](https://github.com/textext/textext/files/2453297/asktext.py.patch.txt)

jcwinkler:
Thanks @veltsov . I will review this during the week.

jcwinkler:
Should we provide this as a checkbox? I think so.
Should word wrap be a default behaviour? It is a matter of taste. Would not set it as default. User can select it and TexText remembers the state. On a per node basis or on a per machine basis?

leberwurstsaft:
I believe it should be a per user/machine setting.

jcwinkler:
**Word Wrap OFF:**
![grafik](https://user-images.githubusercontent.com/11866252/49169343-db232880-f339-11e8-823e-315e56b8d223.png)

**Word Wrap ON:**
![grafik](https://user-images.githubusercontent.com/11866252/49169474-1d4c6a00-f33a-11e8-899c-4fba269cf8d2.png)

Without GTKSourceView there will be only one entry in the Menu. TK is open.

Any suggestions w.r.t. to the position of the checkbox?

sizmailov:
All looks good to me. The only missing part is storage of values in config. It holds for all menu items, not only "wrap" option. Or I missed something?

PR #79, opened 2018-12-01
Feature/word wrap

jcwinkler:
Implements word wrap feature incl. storage of config, see issue #47 

Some major changes were necessary in `asktext.py` to ensure that this feature is also available in GTK version of GUI without GTKSourceView.

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -125,6 +125,7 @@ try:
 
         DEFAULT_ALIGNMENT = "middle center"
         DEFAULT_TEXCMD = "pdflatex"
+        DEFAULT_GUI_WORDWRAP = False
 
         def __init__(self):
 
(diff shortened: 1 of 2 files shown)


jcwinkler:
GTK GUI -> see issue #47 

TK version:

grafik

grafik

sizmailov:
Found two bugs: 
 - The "word wrap" setting is stored/read properly, but it does not affect actual word wrap, see screenshot 
 - Duplicate of word wrap checkbox appears in UI, should be just one in `View` menu

image

jcwinkler:
> The "word wrap" setting is stored/read properly, but it does not affect actual word wrap, see screenshot

Fixed.

> Duplicate of word wrap checkbox appears in UI, should be just one in `View` menu

I did this intentionally to offer a "quick access".

sizmailov:
> I did this intentionally to offer a "quick access".

It's not a good idea, simply because the need of switch is too infrequent. You will probably do it once per installation. 
At the same time it takes valuable space.

jcwinkler:
After one week of work with the modified extension I have to admit that it is enough to have the word wrap switch in the menu. I removed the checkbox.

Issue #48, opened 2018-10-15
Your LaTeX code has problems


JohnCovering:
Hello,

I tried to run TexText 0.8.1 with Inkscape 0.92.3 on Windows 10, but I got the following error when clicking on the Save button.
The LaTeX code was : $x$

With TexText 0.7.3 I don't have any error.

Your LaTeX code has problems:

Traceback (most recent call last):
  File "asktext.py", line 599, in cb_ok
    self._texcmd_cbox.get_active_text().lower())
  File "textext.py", line 292, in callback
    tex_cmd, original_scale=current_scale)
  File "textext.py", line 393, in do_convert
    new_svg_ele = converter.convert(text, preamble_file, scale_factor, tex_command)
  File "textext.py", line 820, in convert
    self.tex_to_pdf(tex_command, latex_text, preamble_file)
  File "textext.py", line 763, in tex_to_pdf
    raise RuntimeError("Your LaTeX code has problems:\n\n{errors}".format(errors=parsed_log))
RuntimeError: Your LaTeX code has problems:

sizmailov:
I wonder is your error message truncated? Is there anything after `RuntimeError: Your LaTeX code has problems:`?

JohnCovering:
No, there's nothing after "RuntimeError: Your LaTeX code has problems:"

michaelhaar:
I am getting the same error message.

- operating system: Win10 64bit
- inscape version: 0.92.3 with Python 2.7
- Miktex version 2.9.6821
- installed PyGTK-2.24-Inkscape-0.92.3-64bit.exe, pstoedit Version 3.73, ghostscript Version 9.25 and ImageMagick-7.0.8-14-Q16-x64-static.exe as described in the wiki

I also tried to use install textext on my office computer. It also displays the same error message.

Note: The error only occurs if `pdflatex` is selected. `xelatex` works fine 😄

michaelhaar:
@sizmailov TexText version is 0.8.1.

The error doesn't depend on the text itself. It always gives the same error

sizmailov:
Can you create a tex file with following content and compile it from command line with pdflatex? 

one.tex:
```latex
\documentclass{article}

\usepackage{amsmath,amsthm,amssymb,amsfonts}
\usepackage{color}

\pagestyle{empty}
\begin{document}
% ***Your code***
\end{document}
```

two.tex:
```latex
\documentclass{article}

\pagestyle{empty}
\begin{document}
% ***Your code***
\end{document}
```

[...]

PR #84, opened 2018-12-04
Feature/show full log

### LINKED ISSUE #77: Provide full latex output on error
Sometimes latex log parser fails. Full log would be helpful in such cases.

Let's show full latex log if:
  1. latex command fails 
  2. parser fails or finds nothing

It would be default behavior and should cover almost all scenarios. 

Full log still might be helpful even if parser finds errors, lets have a menu item option
`Show latex output` with options:  `Parsed` and `Raw`. 
If `Raw` selected, parser results would be ignored.

sizmailov:
Stdout is hidden (default)
image

Stdout expanded

image

Stderr is empty so not shown 

close #77

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -40,7 +40,7 @@ Florent Becker and Vladislav Gavryusev for contributions.
 .. _InkLaTeX: http://www.kono.cis.iwate-u.ac.jp/~arakit/inkscape/inklatex.html
 """
 
-__version__ = "0.9.0-dev"
+__version__ = "0.9.1-dev"
 __docformat__ = "restructuredtext en"
 
 import abc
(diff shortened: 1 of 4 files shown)


sizmailov:
Updated error dialog:


Default:

image

Expanded and resized

image

jcwinkler:
I reproduced the problem reported in issue #48 (missing package mptopdf and missing permissions for installation on the fly). This issue is perfectly handled by the feature commited by @sizmailov.

**Situation in TexText 0.8.1:**
grafik

**Situation now, after failed compilation**
grafik

**After expanding stderr**
grafik

Unfortunately, the refered log file does not contain helpful information:
```
2018-12-07 15:11:55,227+0100 INFO  pdflatex - starting with command line: "C:\Program Files\MiKTeX 2.9\miktex\bin\x64\pdflatex.exe" tmp.tex -interaction=nonstopmode -halt-on-error
2018-12-07 15:11:55,227+0100 INFO  pdflatex - allowing known shell commands
2018-12-07 15:11:55,227+0100 INFO  pdflatex - enabling input (output) from (to) processes
2018-12-07 15:11:55,337+0100 FATAL pdflatex - GUI framework cannot be initialized.
2018-12-07 15:11:55,337+0100 FATAL pdflatex - Info: 
2018-12-07 15:11:55,337+0100 FATAL pdflatex - Source: Libraries\MiKTeX\UI\Qt\mikuiqt.cpp
2018-12-07 15:11:55,337+0100 FATAL pdflatex - Line: 73
2018-12-07 15:11:55,337+0100 INFO  pdflatex - finishing with exit code 1
```
Only the hint to the GUI framework shows (to the experienced) user that MiKTeX tried to open its package installation window.

I think we can provide a solution in the FAQ for this case, what do you think @sizmailov ? Or allow MiKTeX to spawn its window, but I think it is a little bit too much to provide such stuff for all available TeX distros...

sizmailov:
The source of problem - missing package. And it should be reported, nothing beyond that. 
We should either 
1) prohibit installation of packages on the fly via command line option if such option exists (then log parser should report about missing package automatically) 
or 
2) tweak log parser to produce meaningful error message in this case

sizmailov:
In addition for can add check for default packages in requirements as you suggested https://github.com/textext/textext/issues/48#issuecomment-443175456

But general procedure to handle missing packages errors is still required.

PR #88, opened 2018-12-09
fix(preview): limit height of preview image

sizmailov:
Preview image is limited in width, but not in height. Tall preview images inadequately shrink input text area and resize TexText window even beyond display size. 

This PR limits preview image height to 150 px. 

image


jcwinkler:
Good improvement, thank you!

PR #89, opened 2018-12-09
Feature/store gui settings

jcwinkler:
Settings for auto-indent, white-spaces, line-numbers and tabwidth are stored in config, too (see remark in issue #47) 

Use dictionary for passing the options between AskText instances and TexText core, `gui_` is the corresponding prefix in the related keys. Simple, but hopefully enough for our purpose.

I also gave the Tk interface a slightly more compact appearance by reducing the default line number of the TextView widget:
grafik

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -290,7 +290,7 @@ try:
                 else:
                     logger.debug("Using default tex converter `%s` " % current_tex_command)
 
-                gui_wordwrap = self.config.get("gui_wordwrap", TexText.DEFAULT_GUI_WORDWRAP)
+                gui_config = {"word_wrap": self.config.get("gui_wordwrap", TexText.DEFAULT_GUI_WORDWRAP)}
 
                 # Ask for TeX code
                 if self.options.text is None:
@@ -322,7 +322,7 @@ try:
                                                  current_alignment=alignment, current_texcmd=current_tex_command,
                                                  tex_commands=sorted(list(
                                                      self.requirements_checker.available_tex_to_pdf_converters.keys())),
-                                                 word_wrap=gui_wordwrap)
+                                                 gui_config=gui_config)
 
                     def save_callback(_text, _preamble, _scale, alignment=TexText.DEFAULT_ALIGNMENT,
                                  tex_cmd=TexText.DEFAULT_TEXCMD):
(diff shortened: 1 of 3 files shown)


extension/textext/__init__.py
```diff
@@ -290,7 +289,8 @@ def effect(self):
                 else:
                     logger.debug("Using default tex converter `%s` " % current_tex_command)
 
-                gui_wordwrap = self.config.get("gui_wordwrap", TexText.DEFAULT_GUI_WORDWRAP)
+                # GUI related settings start with "gui_"
+                gui_config = {k.split("gui_")[1]:v for k,v in self.config.items() if k.startswith("gui_") and len(k) > 4}
```
sizmailov: Let's have separate "gui" item in config

```python
gui_config = self.config.get("gui",{})
```

extension/textext/__init__.py
```diff
@@ -338,10 +338,12 @@ def preview_callback(_text, _preamble, _preview_callback, _tex_command):
                                                     _tex_command)
 
                     with logger.debug("Run TexText GUI"):
-                        _, _, _, gui_wordwrap = asker.ask(save_callback, preview_callback)
+                        gui_config = asker.ask(save_callback, preview_callback)
```
sizmailov: Now ask returns something meaningful :+1:

extension/textext/__init__.py
```diff
@@ -338,10 +338,12 @@ def preview_callback(_text, _preamble, _preview_callback, _tex_command):
...
 
                     with logger.debug("Saving global GUI settings"):
-                        self.config["gui_wordwrap"] = gui_wordwrap
+                        for key, value in gui_config.items():
```
sizmailov: Let's store whole `gui_config` object as `gui`

```python
self.config["gui"] = gui_config
```

extension/textext/utility.py
```diff
@@ -169,6 +169,9 @@ def get(self, key, default=None):
             return default
         return result
 
+    def items(self):
```
sizmailov: This function will be unused

sizmailov:
Tested on Linux with PyGTK. Works as expected. :+1:

extension/textext/__init__.py
jcwinkler: I'm always amazed about what's possible... :smiley: I changed it in that way.

PR #93, opened 2018-12-12
Feature/add scroll preview mode

sizmailov:
Sometime output of TexText is too big for scaled preview. 

The PR add "scaled" mode to zoom in and see fine details of generated image. 

To switch between preview modes double click on preview area.

The default view is always "scaled". Don't think we need to store this option anywhere. 

image

image

@jcwinkler Tool-tip messages are not clear now and should be reworded to smth. like  "Double click to zoom in/out", what do you think?

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -774,6 +774,7 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
 
                 self._preview.set_from_pixbuf(pixbuf)
                 self._preview.set_size_request(pixbuf.get_width(), pixbuf.get_height())
+                self._preview_scroll_window.set_tooltip_text("Double click to switch to scroll mode")
 
                 return image_height
 
@@ -788,7 +789,8 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
                     desired_preview_area_height += scroll_bar_width
 
                 self._preview.set_from_pixbuf(self._pixbuf)
-
+                self._preview.set_size_request(image_width, image_height)
+                self._preview_scroll_window.set_tooltip_text("Double click to switch to scaled mode")
                 return desired_preview_area_height
 
             if self.preview_representation == "SCROLL":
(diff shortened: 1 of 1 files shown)


jcwinkler:
I suggest 
- `Double click to see complete output scaled to window width` (when output is in original size)
- `Double click to see output in original size` (when output is scaled)

sizmailov:
Thanks for feedback! I like slightly shorter version:
- `Double click: scale to fit window` (height is also limited, so we fitting to  `window_width x 150px`. The message is a bit imprecise, but I think "window"  is better than "preview area" or smth. wordy)
- `Double click: scale to original size`

sizmailov:
I also want to suppress switch between modes if they appear the same way. It would be the case (quite popular, I think) for a small input. Additional meaningless controls will not bother users.

sizmailov:
With big preview image TeXText window has minimum height ~690 px. It should fine for most of users (https://www.w3schools.com/browsers/browsers_display.asp)

I think we can save some space putting "Preamble File" and "TeX command" at same line, but it another task.

Here is minimum TexText window with big preview image: 
 
image


I think this PR is ready to merge.

jcwinkler:
@sizmailov On Windows I get this output when showing the complete image. Seems a whole page is rendered.

grafik

SVG node in Inkscape document has correct bbox.

sizmailov:
It's really wired. 

Why preview has white background? 

The whole page should have identical margins, but here it's not the case. 

We produce png with `--export-drawing-area` option, so empty space should be there. I'll try to reproduce on windows.

sizmailov:
Ok. It happens only with `pstoedit`on both Linux/Windows.  `pdf2svg` is fine on Linux/Windows. Seems like it's not problem of this particular PR...

sizmailov:
Looks like it `pstoedit` always worked this way. With `imagemagick` we used to `-trim` borders and they did not appear. In importing of resulting svg we are looking for "g" node of `pstoedit` output, so background also disappears. 

I'll try to find a reasonable solution

Any how it's unrelated to current PR.

sizmailov:
Background issue should be fixed with #96

jcwinkler:
Works as expected now on Windows with pdf2svg and pstoedit backend.

PR #94, opened 2018-12-12
Save vertical space

sizmailov:
Save some vertical space by putting "TeX command" and "Preamble file" on the same line

The order is also changed. Now it reads just as command line `pdflatex default_packages.tex`. 

image

It might feel unusual at first, but I think it's matter of habit


leberwurstsaft:
Saving vertical space is always nice. It looks crowded because of the extra lines around the preamble file box. Unfortunately, the "Clear" button would be unclear, if there was no surrounding box.

sizmailov:
@leberwurstsaft good point!

Here is version with a little bit tweaked spacing:

image

Is it satisfactory now? I'll add to PR

leberwurstsaft:
Great!

jcwinkler:
Thank your very much, looks really great now! Nevertheless, I would like to postpone this to a release > 0.9.0 for specific reasons I will explain later.

sizmailov:
OK, no problem. Not a big deal since 0.9.0 should appear shortly, in a week or two I think.

sizmailov:
Big alignment icons make whole UI bigger, because of rows "normalization". I reduced size of icons and tuned them for sharp 21x21 pixel render:

image

jcwinkler:
Thank you, looks very good.

Although I did not want to change the preamble line (I wanted to add something to edit the tex file there) I will merge this now because of #99 which eventually will lead to major modifications in the GUI anyway.

PR #95, opened 2018-12-12
Refactor/replace md with rst

sizmailov:
README.rst is specical and required to be properly rendered by github. It leaves some redundancy. 

CHANGELOG, AUTHORS and HISTORY is rewritten as `.rst` and reused in docs. 

`docs/index.html` has two more tables of content: "Credits" and "Changelog"

New .rst role "bb_issue_num" and "bb_issue" introduced to automatically create links to bitbucket issues. 

README.rst is rewritten to match docs/index, but it's not the final version, you may add extra stuff in it. 

All doc links are stable (i.e. they would not change with doc section name change) 

Note: (installation-toc) and (usage-toc) anchors are here but not deployed to textext.github.io yet, so you will get to top of the page.


jcwinkler:
Thanks! Exactly what I wanted!

PR #97, opened 2018-12-13
feat(ui): Add alignment icons

sizmailov:
Replace alignment anchor position text labels with icons

image


jcwinkler:
It does not look quite nice on Windows:

grafik

Problem is this little down arrow which, why the hell, is always placed in the center of the combo boxes on Windows. Also, we have three "heights" now: The in the preamble file/ TeX command, the alignment buttons and the preview/save/cancel buttons. If you have the time, maybe you can "normalize" this a bit? Maybe you can take the distance tweaks from pr #94 and then adjust the button sizes ?

sizmailov:
Good news: there is "style property" named `GtkComboBox::arrow-size` ([doc](https://developer.gnome.org/pygtk/stable/class-gtkcombobox.html#style-properties-gtkcombobox)) . 

Bad news: I didn't get it to work. 

Seems like next code should exterminate arrows in all combo boxes, but it doesn't. 

```python
    rc_str = """
               style 'no_arrow_style' {
                   GtkComboBox::arrow-size = 0
               }
               widget_class '*' style 'no_arrow_style'
               """
    gtk.rc_parse_string(rc_str)
```

I have no idea..

sizmailov:
Tested on Linux and Windows

jcwinkler:
@sizmailov This goes into 0.9.1. Thank you!

PR #100, opened 2018-12-13
Feature/normalize ui rows height

sizmailov:
Three rows of controls are normalized to same height. Row with Cancel/Preview/Save buttons is NOT resized. I find it decoupled from other three and find it visually more attractive. 

Horizontal spacing is also changed, similarly to #94 

Cancel/Preview/Save buttons row is resized:
image

Cancel/Preview/Save buttons row is NOT resized:
image


jcwinkler:
Well done, thanks!

PR #102, opened 2018-12-21
Added issue templates

jcwinkler:
There will be some double information since system information is also found in the log but I think it is worth to see what the user "thinks" he has installed compared to what TexText finds. Also there may be some users who do not find the log file (extension is in hidden directories on all platforms...)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,44 @@
+---
+name: Bug report
+about: Create a report to help us improve the extension
+title: ''
+labels: ''
+assignees: ''
+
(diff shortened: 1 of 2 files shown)


sizmailov:
Sometimes output is quite large. I think we should put all of them under spoiler:

<details>
<summary>Output log:</summary>

```
<<< REPLACE THIS TEXT WITH ACTUAL OUTPUT>>> 
```
</details>



Code for spoiler template (replace `tripple_backtick` with `` ``` ``:
```
<details>
<summary>Output log:</summary>

tripple_backtick
<<< REPLACE THIS TEXT WITH ACTUAL OUTPUT>>> 
tripple_backtick
</details>
```

PR #116, opened 2019-02-19
feat(ui): Add new node content UI option (PyGTK only)

### LINKED ISSUE #115: Automatically include $$ in input text box (QOL)
**Is your feature request related to a problem? Please describe.**
I've forgotten to do the above more times than I care to recall.

**Describe the solution you'd like**
When creating a new Tex Text, instead of starting with nothing in the input text box, start with $$. Even better if the cursor is automatically between both the $.

**Additional context**
The alternative is to automatically include $$ around the expression if it's not provided at all (assuming there's no reason to use Latex in inkscape if not for the mathematics bit), but that's probably too auto-magical.

My assumption here, based on my own use-case, is that Latex is primarily used to input various mathematical formulae, and easing this would be the goal of this plugin.

sizmailov:
Related issue(s):

resolves #115 

New node can be initialized with 
   - empty content (default)
   - inline math template `$<cursor>$`
   - display math template `$$<cursor>$$`

The choice is controlled by "New node content" GUI option, located at  "Settings" -> "New node content". 

Implemented in PyGTK only. Tk interface is not affected. 

image

Short checklist:
- [x] Tested with Inkscape version: [0.92.3]
- [x] Tested on Linux, Distro: [kubuntu 18.04]

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -195,6 +195,8 @@ class AskText(object):
     DEFAULT_AUTOINDENT = True
     DEFAULT_INSERTSPACES = True
     DEFAULT_TABWIDTH = 4
+    DEFAULT_NEW_NODE_CONTENT = "Empty"
+    NEW_NODE_CONTENT = ["Empty", "InlineMath", "DisplayMath"]
 
     def __init__(self, version_str, text, preamble_file, global_scale_factor, current_scale_factor, current_alignment,
                  current_texcmd, tex_commands, gui_config):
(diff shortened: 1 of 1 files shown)


jcwinkler:
Thanks, looks good to me at a first glance and I will test it the next days.

sizmailov:
The only inconvenience I experienced so far is that TexText window doesn't get focus on launch. If I click on input text area the cursor eventually goes at the end, so I have to move it back in between of `$$` to start typing. TexText window started from console without inkscape does not have focus either. 

Focusing issue needs to be fixed to let this feature shine bright.

jcwinkler:
- [x] Tested on Windows, Version: [10 1803 and 8.1]

jcwinkler:
> The only inconvenience I experienced so far is that TexText window doesn't get focus on launch. If I click on input text area the cursor eventually goes at the end, so I have to move it back in between of `$$` to start typing.

I cannot reproduce this behavior on Windows. When creating a new node the window is active and the cursor is placed between the $'s.

sizmailov:
> I cannot reproduce this behavior on Windows. 

Good, hopefully it's just me. Currently I have no idea what went wrong. I'll investigate it once I have some spare time. 

Also, we need to add corresponding section to docs.

jcwinkler:
> Also, we need to add corresponding section to docs.

I am currently writing on the docs.

Issue #122, opened 2019-04-16
Migrate from GTK+ 2.x to GTK+ 3


sizmailov:
Switch to new major version of GTK is part of #107

jcwinkler:
Will be done for Inkscape 1.0 release, see https://github.com/textext/textext/projects/1

Issue #107, opened 2019-01-21
Feature request: Port to 1.0 and submit textex to Inkscape's stock extensions


Esokrates:
It would be great if you could follow
http://wiki.inkscape.org/wiki/index.php/Updating_your_Extension_for_1.0
and propose the extension to be part of the stock extensions.

Another thing I would wish for is the dummy window to disappear that shows up before the textext dialog appears. If you want a different issue to track this, no problem.

sizmailov:
TexText has extra dependencies that prohibits its inclusion in stock extensions, IMO.

I agree dummy window is annoying, but I think there is no way to avoid it without modifying inkscape source code.

sizmailov:
I [asked](https://gitlab.com/inkscape/inkscape/issues/156#note_174623053) a number of questions about inkscape 1.0 to clarify what needs to be done to support it. 

While GTK3 is required by inkscape the *python bindings* are NOT guaranteed to be installed in the interpreter. So we need to supply GTK3 installers/instructions. 

The extension is required to be python2/3-complaint. Which is a bit sad.

Here is short check-list: 
 - [X] Update GUI to GTK3
 - [ ] Adjust code for new `inkex.py` API
 - [ ] Remove `pdf->svg` dependency
 - [ ] Make extension python3-complaint

sizmailov:
I filed an issue about dummy window https://gitlab.com/inkscape/inkscape/issues/267 . Maybe we will be able to get rid of it in near future)

sizmailov:
It turns out that the required change to suppress dummy window was rather small. Hopefully it's going to be [merged](https://gitlab.com/inkscape/inkscape/merge_requests/698) in  inkscape 1.0 release.

sizmailov:
:tada: :tada: :tada:
Inkscape 1.0 will hide "extension is working" window for extensions with custom GUI!

leberwurstsaft:
That's so sweet!

sizmailov:
Good news everyone! We can get rid of "extension is working" window in inkscape<1.0! See #138

jcwinkler:
> Good news everyone! We can get rid of "extension is working" window in inkscape<1.0!

Thanks for figuring this out! This will be in release 0.11!

jcwinkler:
See https://github.com/textext/textext/projects/1

PR #123, opened 2019-04-16
Migration to GTK3+

sizmailov:
Related issue(s): #122 #107 

Replace GTK2 with GTK3

Short checklist:
- [x] Tested with Inkscape version: [0.92.3]
- [x] Tested on Linux, Distro: [Ubuntu 18.04.1]

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -43,13 +43,17 @@ from errors import TexTextCommandFailed, TexTextConversionError
 #   If unsuccessful, try TK
 #   When not even TK could be imported, abort with error message
 try:
-    import pygtk
+    import gi
 
-    pygtk.require('2.0')
-    import gtk
+    gi.require_version("Gtk", "3.0")
+    from gi.repository import Gtk
+
+    from gi.repository import Gdk, GdkPixbuf
 
     try:
-        import gtksourceview2
+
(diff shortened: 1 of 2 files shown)


sizmailov:
I think I changed almost all gtk-involved parts, but I'll revise them once again later.

Almost all changes are one-to-one, but there is a number of deprecated interfaces I replaced. I'll review all the changes in detail a bit later.  

Here is current GTK-3 look of TexText on my system:
image

sizmailov:
I think I can turn this draft to PR shortly

@jcwinkler  The only missing part - is the drop-down arrow for alignment icons. As you found it renders on top of icon on Windows. Old way of removing it definitely doesn't work anymore. We need to check whether change is required for GTK3.

jcwinkler:
Thanks, @sizmailov . I am currently working on the corresponding PyGTK package for Windows. It tuns out that it is not that easy as expected due to tons of crucial depndencies but it should be only a matter of time to get rid of it...

jcwinkler:
grafik

[x] Tested on Windows 8.1, Inkscape 0.92.4

Arrow seems to be OK, too

jcwinkler:
Thanks again for this contribution @sizmailov ! I keep this in a feature branch until the PyGTK-Installer is finalized and reduced to a smart size (currently it's sumo :worried:).

Issue #126, opened 2019-05-04
Incompatibility between Ghostsript 9.27 and pstoedit (all versions)


sindzicat:
Hello!

I'm on Windows 8.1x64.

I created `test.tex`:

```latex
\documentclass{article}
\usepackage{amsmath,amsthm,amssymb,amsfonts}
\usepackage{color}
\pagestyle{empty}
\begin{document}
$\sin(x^2) + \cos(x^2) = R^2$
\end{document}
```

Then compiled it. I tried `xelatex test.tex` and `pdflatex test.tex -interaction=nonstopmode -halt-on-error`. I got a pdf file. This file was fine and had the correct formula.

Then I tried:
```
"C:\Program Files\pstoedit\pstoedit.exe" -f plot-svg test.pdf test.svg -dt -ssp -psarg -r9600x9600 -pta
```

I got the following error:
```
pstoedit: version 3.73 / DLL interface 108 (built: Jun  3 2018 - release build - MS VC++ 1912 - 64-bit) : Copyright (C) 1993 - 2018 Wolfgang Glunz
Error: /undefined in .makeoperator
Operand stack:
   false   rectfill   rectfill   --nostringval--
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1
%stopped_push   1991   1   3   %oparray_pop   1990   1   3   %oparray_pop   1978   1   3   %oparray_pop   1833   1   3   %oparray_pop   --nostringval--   %errorexec_pop
 .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--
Dictionary stack:
   --dict:1033/1123(G)--   --dict:0/20(G)--   --dict:216/300(L)--   --dict:1033/1123(G)--
Current allocation mode is global
Current file position is 15548
GPL Ghostscript 9.27: Unrecoverable error, exit code 1
PostScript/PDF Interpreter finished. Return status 1 executed command : C:\Program Files\gs\gs9.27\bin\gsdll64.dll -q -dDELAYBIND -dWRITESYSTEMDICT -dNODISPLAY -dNOEPS -r9600x9600 C:/Users/SINDZI~1/AppData/Local/Temp/psin2
The interpreter seems to have failed, cannot proceed !
```

How can I fix this?

Best regards,
Andrey.

sindzicat:
[...]

Issue #125, opened 2019-04-24
Mac OS: Where is tmp.svg created? 


osbama:
My tex->pdf->svg toolchain works perfectly both with the recommended text and in my daily use, however, textext does not work, since 

Command "/Applications/Inkscape.app/Contents/Resources/bin/inkscape -f tmp.svg --export-png tmp.png --export-area-drawing --export-dpi=200" 

fails with:  "Can't open file: tmp.svg (doesn't exist)" 

yes, neither can I find tmp.svg. Where does this file get created?

sizmailov:
All data is created in temp directory, usually it's `/tmp/<temp dir>`, but it's deleted once it's no longer needed even in case of error. 

Did you try to manually reproduce textext toolchain https://textext.github.io/textext/usage/troubleshooting.html#trouble-manual-compile ? 

Please copy-paste manual testing results from terminal as markdown code block (enclose in triple back quotes  ```)

PR #136, opened 2019-06-12
Compatibility with pstoedit 3.74 and gs 9.27

jcwinkler:
Related issue: #126 

Precise description of the changes proposed in the pull request: Make TexText compatible with pstoedit 3.74 and gs 9.27. Some harsh adjustments in the tests have been necessary, however, after manually checking the results it looks really good.

Short checklist:
- [x] Tested with Inkscape version: 0.92.3, 0.92.4
- [x] Tested on Linux, Distro: Kubuntu 18.04
- [x] Tested on Windows 8.1 x64


pytests/test_compatibility.py
```diff
@@ -10,16 +10,18 @@
...
     def __enter__(self):
-        self.__name = tempfile.mkdtemp()
+        self.__name = tempfile.mkdtemp(suffix=self.__suffix)
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if self.__name is not None:
-            shutil.rmtree(self.__name)
+            pass
+            #shutil.rmtree(self.__name)
```
sizmailov: Remove debugging stuff.

pytests/test_compatibility.py
```diff
@@ -10,16 +10,18 @@
 
 
 class TempDirectory(object):
-    def __init__(self):
+    def __init__(self, suffix=""):
```
sizmailov: This change feels wrong. Temp dir should be ... temporal, so the name does not matter. If you want to distinguish between test in debug - create root temp directory with named directories inside. While this ad-hoc debug technique might be helpful, I think it should not be merged.

sizmailov:
So pstoedit works with gs?! Great!

pytests/test_compatibility.py
jcwinkler: I changed it such that one can decide whether to have temporary files in the system's temp dir (i.e. the old behavior) or to keep the results in a separate folder (for manual inspection after the tests).

sizmailov: It resides in tests only, so let it be.

sizmailov: Just a thought. We have TempDir guard in main extension code too. I think switching temp dir to known place might be helpful feature in debugging on user's side, note #125.

PR #137, opened 2019-06-16
Possibility to check for gs/pstoedit versions by specifying an operator

jcwinkler:
Enables one to check for e.g. pstoedit versions < 3.74

Related issue: #126

### DIFF
--- a/extension/textext/requirements_check.py
+++ b/extension/textext/requirements_check.py
@@ -4,6 +4,9 @@ import os
 import re
 import subprocess
 import sys
+import operator
+
+opstr = {operator.eq:"=", operator.lt:"<", operator.le:"<=", operator.gt:">", operator.ge:">="}
 
 
 class Defaults(object):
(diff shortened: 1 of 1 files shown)


jcwinkler:
grafik

extension/textext/requirements_check.py
```diff
@@ -572,7 +575,7 @@ def find_tkinter(self):
 
         return RequirementCheckResult(True, ["TkInter is found"])
 
-    def find_ghostscript(self, version=None):
+    def find_ghostscript(self, version=None, op=operator.eq):
```
sizmailov: Instead of adding extra parameter I would change `version` to `version_spec` and require it to be PEP440 complaint. 

Usage example `self.find_ghostscript(">1.0,<2.0")`. 

But since the `requirement_check` is going to be removed along with port to 1.0, I don't think we should spend extra time on it

jcwinkler: I know what you mean. Thank you! Since this will be used only for a limited amount of time I will keep it and modify it later as an exercise :memo: for me.

PR #138, opened 2019-06-19
Remove `extension is working` window for inkscape<1.0

sizmailov:
As It turns out, there was an undocumented option to hide "extension is working" window. 

- [x] Tested with Inkscape version: Inkscape 0.92.3 (2405546, 2018-03-11)
- [x] Tested on Linux, Distro: Ubuntu 18.04

### DIFF
--- a/extension/textext.inx
+++ b/extension/textext.inx
@@ -13,4 +13,5 @@
   <script>
     <command reldir="extensions" interpreter="python">textext/__init__.py</command>
   </script>
+  <options silent="true" />
 </inkscape-extension>


sizmailov:
@jcwinkler It's small change, but it has big affect to every day TexText experience, I think it should be merged to master ASAP

jcwinkler:
I cannot see any changes in the behavior under Windows: still two Windows: TexText and the well known empty small Window.

grafik

But if it improves the situation under Linux I will merge it right now and make a release quickly.

sizmailov:
I don't see why it's not working on windows. Going to check it on virtual machine

jcwinkler:
Ah, I see. I needed to restart Inkscape. I first checked the behavior using old TexText version, then updated TexText and reopend the GUI then still showing the old behavior. But after restarting Inkscape the second Window does not appear anymore. Interesting & sorry for the noise!

sizmailov:
Ok, right. `.inx` files are parsed only once, at initialization, restart of inkscape is required.

Issue #127, opened 2019-05-13
Avoid losing work upon pressing escape


tmoemke:
When using TexText, I frequently lose my text by pressing Esc.

Solutions: (i) disable the shortcut (ii) or add a dialog in case that text was changed.

jcwinkler:
I think this is a good suggestion which I will implement.

sizmailov:
I think new "Settings" entry `Close TexText shortcut` with `Escape|Ctrl+Q|None` options should be enough.

PR #139, opened 2019-06-21
Window close confirmation (configurable)

jcwinkler:
Relates to issue #127:
- Default close shortcut is ESC
- Confirmation for closing TT window when text has been changed
- Confirmation can be disabled

Remark: Maybe it would be "perfect" when also changes in scale factor etc. are detected. But due to lack of time I postpone this...

Short checklist:
- [x] Tested with Inkscape version: 0.92.3, 0.92.4
- [x] Tested on Linux Kubuntu 18.04
- [x] Tested on Windows 8.1 x64

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -531,6 +531,11 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
                  'Wrap long lines in editor to avoid horizontal scrolling', self.word_wrap_toggled_cb)
             ]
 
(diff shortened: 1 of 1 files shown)


sizmailov:
Cool! Works just as expected!

I create a PR to your branch with a bit of refinement of message dialog. I don't like "Yes/No" options on exit dialogs because it requires me to read the dialog message. It's better to spell intentions explicitly, so dialog message can be skipped.

PR #143, opened 2019-09-04
Feature/travis deploy for 1.0 (Stage 1 of 2)

jcwinkler:
This pull request is the first of two for re-enabling travis ci tests.

In this stage an app image of the alpha2 version of Inkscape is installed in the build machine and the installation procedure is tested. The next stage will re-enable the compatibility tests (or at least introduce a simple functional test).

Note: I wanted to have Inkscape 0.92.x and 1.0alpha2 in parallel on my test-machine. Hence I har-coded the alpha2 name and some places which is not very elegant but maybe OK for this test phase...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,15 +29,15 @@ matrix:
 before_script:
    # Deactivated until tests are rewritten
    #- python -m pip install pytest lxml Pillow
-   - wget -c https://inkscape.org/de/gallery/item/14204/Inkscape-883c7bc-x86_64.AppImage
-   - chmod +x Inkscape-883c7bc-x86_64.AppImage
-   - sudo mv Inkscape-883c7bc-x86_64.AppImage /usr/local/bin
+   - wget -c https://inkscape.org/de/gallery/item/14204/Inkscape-883c7bc-x86_64.AppImage -O ~/bin/inkscape.alpha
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,52 +1,44 @@
...
 
 before_script:
-   - python2 -m pip install pytest lxml Pillow
-   - python2 test_installation_script.py 2> /dev/null
-   - python2 setup.py
+   # Deactivated until tests are rewritten
+   #- python -m pip install pytest lxml Pillow
+   - wget -c https://inkscape.org/de/gallery/item/14204/Inkscape-883c7bc-x86_64.AppImage
+   - chmod +x Inkscape-883c7bc-x86_64.AppImage
+   - sudo mv Inkscape-883c7bc-x86_64.AppImage /usr/local/bin
```
sizmailov: 1) `sudo` is not required here. Executable can be placed in `$HOME/bin`. Maybe `$PATH` should be updated as well.

2) I would rename `Inkscape-883c7bc-x86_64.AppImage` to `inkscape.alpha`

jcwinkler: Thanks! Indeed $PATH needs to be updated so requirements.py finds the inkscape.alpha executable.

Issue #148, opened 2019-10-26
Gradient fills not properly rendered


ouening:
**- Problem Description **
When I use textext to draw some pictures using pgfplots package, some error happens. Detailed description is listed bellow:
1. Firstly I modified the file `default_packages.tex` to support pgfplots:
```
\usepackage{amsmath,amsthm,amssymb,amsfonts}
\usepackage{color}
\usepackage{times}
\usepackage{epsfig}
\usepackage{graphicx}
\usepackage{amsmath}
\usepackage[lined,boxed,commentsnumbered, ruled]{algorithm2e}
\usepackage{bm}
\usepackage{multirow}
\usepackage{pgfplots}
\pgfplotsset{compat=1.16}
\usepackage{pgfplotstable}
\usepackage{verbatim}
\usepgfplotslibrary{patchplots}
\usepackage{tikz}
\usetikzlibrary{mindmap,shadows}
\usepackage[hidelinks,pdfencoding=auto]{hyperref}
\usepackage{smartdiagram}
\usepackage{fullpage}
\usepackage{fourier}
\usepgfplotslibrary{colorbrewer}
\usepackage{sansmath}
\usetikzlibrary{shadings,intersections}
```
2. Problem 1: No effect using **colorbar**, the code is:
```
\begin{tikzpicture}
\begin{axis}[
title={$x \exp(-x^2-y^2)$},
xlabel=$x$, ylabel=$y$,
colorbar horizontal,
small,
]
\addplot3 [
surf,
domain=-2:2,
domain y=-1.3:1.3,
] {exp(-x^2-y^2)*x};
\end{axis}
\end{tikzpicture}
```
the result is:
![no-colorbar](https://user-images.githubusercontent.com/15122811/67614995-fa039680-f7f8-11e9-9d9e-f757890f26a8.png)
As we can see that the colorbar has nothing, but no any warning or error happens.

3. Problem 2: Can't launch 3rd party executables (using contour gnuplot), I have installed gnuplot and add it to system paths, the code I used is:
```
\begin{tikzpicture}
\begin{axis}[
title={$x \exp(-x^2-y^2)$},
enlarge x limits,
view={0}{90},
xlabel=$x$, ylabel=$y$,
small,
]
\addplot3[
domain=-2:2,
domain y=-1.3:1.3,
contour gnuplot={number=14},
thick,
] {exp(-x^2-y^2)*x};
\end{axis}
\end{tikzpicture}
```
Error happens:
```
[...]

PR #149, opened 2019-11-01
Support `<defs>`

sizmailov:
Related issue(s):
   #148 

  - fix interactive install with python3
  - embed `<defs>` tag in textext object

An alternative implementation with populating root `<defs>` would result to buggy copy to another document and potential pollution of root `<defs>`. 

Currently I'm not satisfied with id replacement performance. It takes 8 seconds on my machine on  first example from #148, therefore I mark the PR as a draft for now.

Short checklist:
- [x] Tested with Inkscape version: [ Inkscape 1.0beta1 (e7e369847f, 2019-11-25) ]
- [x] Tested on Linux, Distro: [ Ubuntu 18.04.1 LTS ]

**Edit:** update tested inkscape version

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -47,6 +47,7 @@ import hashlib
 import logging
 import logging.handlers
 import math
+import re
 import os
 import platform
 import sys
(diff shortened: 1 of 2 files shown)


sizmailov:
Improved performance for id replacement by using re module.
Now replacement in example #148 takes 35 ms instead of 8 seconds (200+ times fold!)

@jcwinkler Please review and test
@ouening Please test this branch as well

jcwinkler:
I tested it under Windows with the example from #148. The gradient color bar is shown now but I think the color distribution is not correct.

When compiling natively with pdflatex one obtains:

grafik

When compiling via TexText in Inkscape 1.0beta one obtains:

grafik

Quite strange.

sizmailov:
@jcwinkler Good finding! Just test on linux - no problem. I'm not sure I'll be able to test/debug under windows soon :(

jcwinkler:
For me it looks like an Inkscape problem under Windows, Doing the conversion pdf->svg manually by executing

```batch
C:\InkscapeBeta\bin\inkscape.exe --without-gui --pdf-poppler --pdf-page=1 --export-type=svg --export-text-to-path --export-area-drawing --export-file test.svg test.pdf
````

results in the same false colormap. I will file a report there.

sizmailov:
Seems so. Could you report it to inkscape tracker on gitlab?

sizmailov:
By the way which version do you use? I'm using nightly builds, so I'm getting new bugs and fixes ASAP :)

jcwinkler:
> By the way which version do you use?
1.0beta1

Under Linux beta1 it works as expected.

jcwinkler:
https://gitlab.com/inkscape/inbox/issues/1260

jcwinkler:
No response up to now in the Inkscape bug tracker. Since the issue desribed above is not related to this PR I merge it now.

PR #151, opened 2019-11-13
Cleanup executable setting

aerusso:
Related issue(s):
This is one of the two patches in #146.

This code cleanup removes unnecessarily set executable marking, both `#!` and executable bit.

It is unrelated to any Python 3 changes, this is a vestige that was accidentally imported.

Short checklist:
- [x] Tested with Inkscape version: [develop]
- [x] Tested on Linux, Distro: [Debian]


aerusso:
@sizmailov In the other thread, you commented that this patch was out of scope. Would you be willing to endorse this version, in this separate MR?

This is important because some package managers complain when they see executable bits in the wrong places, or referring to binaries that may not exist on that system (like lintian Debian, for example).

sizmailov:
Looks good to me. It's 100% safe to remove `+x` flags and shabangs since these scripts never run as standalone app.

PR #154, opened 2019-11-28
Doc: Install instructions for Inkscape 1.0

jcwinkler:
I would like to provide instructions on how to test current development version of TexText with Inkscape 1.0 beta while Inkscape 0.94.x is also installed and on the system path.

This PR offers instructions for Linux, as I use it. You can see it here: 

https://jcwinkler.github.io/textext/install/linux-beta.html

Windows will follow.

If some more convenient ways are possible please let me know. Thanks!

### DIFF
--- a/docs/source/install/linux-beta.rst
+++ b/docs/source/install/linux-beta.rst
@@ -1,0 +1,100 @@
+.. |TexText| replace:: **TexText for Inkscape 1.0 beta**
+.. |Inkscape| replace:: **Inkscape 1.0 beta**
+
+.. role:: bash(code)
+   :language: bash
+   :class: highlight
+
+.. role:: latex(code)
+   :language: latex
+   :class: highlight
+
+.. _linux-beta-install:
+
+==================
+|TexText| on Linux
+==================
+
+This guide explains how to install and use |TexText| together with |Inkscape| on a system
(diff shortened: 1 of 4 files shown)


docs/source/install/linux-beta.rst
```diff
@@ -0,0 +1,100 @@
...
+
+        chmod u+x Inkscape-fe3e306-x86_64.AppImage
+
+3. Extract the AppImage:
+
+    .. code-block:: bash
+
+        ./Inkscape-fe3e306-x86_64.AppImage --appimage-extract
+
+After this operation |Inkscape| will reside in ``~/squashfs-root/bin``. You can test it by
```
sizmailov: Doesn't it extract to current folder? Do you assume here that CWD in $HOME?

jcwinkler: Good catch. It is extracted into the current directory. Thanks! I update the doc.

jcwinkler:
Here the Windows version, basically the same procedure as on Linux:

https://jcwinkler.github.io/textext/install/windows-beta.html

jcwinkler:
> I would merge this PR to master in order to deploy docs automatically.

@sizmailov I will do some further cleanup in other parts of the code. After additional PR I will merge all this stuff into master while also providing a pre-release.

Issue #134, opened 2019-06-03
live preview


Foadsf:
I know it might be too much to ask, but having live preview could be really helpful. I have listed several examples [here in this post](https://tex.stackexchange.com/a/420307/102825). For example, an interface similar to [KLatexFormula ](https://klatexformula.sourceforge.io/) would be great.

jcwinkler:
Thanks for the suggestion. I think this is a feature which is nice to have but which should not be considered as the most important now. We have to focus our limited resources on the transition to Inkscape 1.0 compatibility. Later we should evaluate how to further improve the extension.

sizmailov:
I think a shortcut for preview will be a reasonable compromise.

PR #159, opened 2019-12-04
Feature/preview shortcut

jcwinkler:
This PR adds a shortcut for the preview. Furthermore, the shortcuts are shown in the tooltip of the corresponding buttons now. Since the can vary for the cancel operation some handling is required.

Related issue(s): #134

Short checklist:
- [x] Tested with Inkscape version: 1.0 beta1 (Python 2.7 and 3.7)
- [x] Tested on Linux, Distro: Kubuntu 18.04
- [x] Tested on Windows, Version: 8.1

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -700,6 +700,9 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
 
         def close_shortcut_cb(self, action, previous_value, sourceview):
             self._gui_config["close_shortcut"] = self.CLOSE_SHORTCUT[action.get_current_value()]
+            self._cancel_button.set_tooltip_text(
+                "Don't save changes ({})".format(self._close_shortcut_actions[action.get_current_value()][2]).replace(
+                    "_", ""))
 
         def confirm_close_toggled_cb(self, action, sourceview):
             self._gui_config["confirm_close"] = action.get_active()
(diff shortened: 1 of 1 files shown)


extension/textext/asktext.py
```diff
@@ -899,20 +907,20 @@ def create_buttons(self):
             button_box.set_spacing(spacing)
 
             self._cancel_button = Gtk.Button(stock=Gtk.STOCK_CANCEL)
-            self._cancel_button.set_tooltip_text("Don't save changes")
+            self._cancel_button.set_tooltip_text("Don't save changes (ESC)")
```
sizmailov: Shouldn't it be formatted same way as above?

extension/textext/asktext.py
```diff
@@ -716,6 +719,11 @@ def cb_key_press(self, widget, event, data=None):
                 self._ok_button.clicked()
                 return True
 
+            # Show/ update Preview shortcut (CTRL+P)
+            if Gdk.keyval_name(event.keyval) == 'p' and Gdk.ModifierType.CONTROL_MASK and event.state:
+                self._preview_button.clicked()
```
sizmailov: Shortcut needs to be disabled for the time preview is being generated, and reactivated again on success/compilation error as well as preview button

extension/textext/asktext.py
```diff
@@ -700,6 +700,9 @@ def new_node_content_cb(self, action, previous_value, sourceview):
 
         def close_shortcut_cb(self, action, previous_value, sourceview):
```
sizmailov: Can we change this vague name to something more meaningful?

sizmailov: Well, it might go to a separate refactor PR

extension/textext/asktext.py
jcwinkler: What do you mean?

extension/textext/asktext.py
jcwinkler: Will provide separate PR for this. Same holds for save button. Also, if they are clicked by the mouse.

extension/textext/asktext.py
sizmailov: ESC is not the only possible option. It might be `Ctrl+Q`, etc

jcwinkler: This is handled here (called when the settings are changed or loaded from config.json):
https://github.com/textext/textext/blob/705e47c39896cdf7a150cbe834daa8fb24b3a1b8/extension/textext/asktext.py#L701-L705

sizmailov: Then this line is useless and should be deleted.

extension/textext/asktext.py
sizmailov: Ok

extension/textext/asktext.py
jcwinkler: Unfortunately not. During dialog creation ``close_shortcut_cb`` is not called when we have the default behavior (ESC).

It is not clean due to historic development of the code. I know.

I will address this in a separate PR,

PR #164, opened 2019-12-18
Hotfix: Illegal string comparison

jcwinkler:
Since Python 3.8 a warning is thrown in constructs like

if x is not 'None'

with x being a string since this is not an comparison between objects. (Python 3.8 is new interpreter in Inkscape beta 2 on Windows...)

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -718,7 +718,7 @@ if TOOLKIT in (GTK, GTKSOURCEVIEW):
 
             # Cancel dialog via shortcut if set by the user
             close_shortcut_value = self._gui_config.get("close_shortcut", self.DEFAULT_CLOSE_SHORTCUT)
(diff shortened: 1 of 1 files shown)


extension/textext/asktext.py
```diff
@@ -718,7 +718,7 @@ def cb_key_press(self, widget, event, data=None):
 
             # Cancel dialog via shortcut if set by the user
             close_shortcut_value = self._gui_config.get("close_shortcut", self.DEFAULT_CLOSE_SHORTCUT)
-            if close_shortcut_value is not 'None':
+            if close_shortcut_value != 'None':
```
sizmailov: This check can be removed completely, right?

jcwinkler: ...indeed :smiley:

sizmailov:
@jcwinkler Did you mean to close?

PR #166, opened 2019-12-18
Hotfix: Scale factor is ignored in new nodes

jcwinkler:
**Before:**

grafik

**After:**

grafik

In both cases scale factor 1.0 is used.

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -424,8 +424,9 @@ try:
                         height = self.svg.unittouu(self.get_document_height())
 
                         x, y, w, h = tt_node.bounding_box()
-                        tt_node.transform = Transform(tt_node.transform) * Transform(translate=(-x + width / 2 - w / 2,
-                                                                                              -y + height / 2 - h / 2))
+                        tt_node.transform = tt_node.transform * \
+                                            Transform(translate=(-x + width / 2 - w / 2, -y + height / 2 - h / 2)) * \
+                                            Transform(scale=scale_factor)
                         tt_node.set_meta('jacobian_sqrt', str(tt_node.get_jacobian_sqrt()))
 
                         self.svg.get_current_layer().add(tt_node)


extension/textext/__init__.py
```diff
@@ -422,8 +422,10 @@ def do_convert(self, text, preamble_file, user_scale_factor, old_svg_ele, alignm
                         height = self.svg.unittouu(self.get_document_height())
 
                         x, y, w, h = tt_node.bounding_box()
-                        tt_node.transform = Transform(tt_node.transform) * Transform(translate=(-x + width / 2 - w / 2,
-                                                                                              -y + height / 2 - h / 2))
+                        nodetransform = Transform(
+                            translate=(-x + width / 2 - w / 2, -y + height / 2 - h / 2)) * Transform(
+                            "scale(%f)" % scale_factor)
+                        tt_node.transform = Transform(tt_node.transform) * nodetransform
```
sizmailov: Style correction:
```python
tt_node.transform = tt_node.transform * \
                    Transform(translate=(-x + width / 2 - w / 2, -y + height / 2 - h / 2)) * \
                    Transform(scale=scale_factor)
 ```

jcwinkler: Scaling was done by the Pdf2SvgConverter invoked here:

https://github.com/textext/textext/blob/29988433eaaf0b147da0e18f14c7689cfea8bab8/extension/textext/__init__.py#L436-L445

and implemented here:
https://github.com/textext/textext/blob/29988433eaaf0b147da0e18f14c7689cfea8bab8/extension/textext/__init__.py#L658-L667

sizmailov: Yeah, right. Then scaling needs to be first, just as in your PR, Good!

sizmailov: Does it mean that other branch (reedited node) is affected as well?

jcwinkler: There, ``tt_node.align_to_node(old_svg_ele, alignment, relative_scale)`` does the job:

https://github.com/textext/textext/blob/3527dae7d70ad41212b5d5dfaaf35ee117723463/extension/textext/__init__.py#L432-L435

sizmailov:
Tested, seems to work now!

PR #170, opened 2019-12-20
[WIP] Hotfix/wrong node placement

jcwinkler:
New nodes are not properly set in the center of the document.

**Situation before the fix**

grafik

**Situation after the fix**

grafik

Two reasons for this behavior:
https://github.com/textext/textext/blob/3f29077b6f440e2937de9c7e04505169e7b9102e/extension/textext/__init__.py#L426

In  fact ``tt_node.bounding_box()`` expands to ``x, w, y, h`` given in pt.

Furthermore
https://github.com/textext/textext/blob/3f29077b6f440e2937de9c7e04505169e7b9102e/extension/textext/__init__.py#L423-L424
returns units in mm (if user has mm for the document unit).

I use new method ``get_center_position()`` provided by Inkscape extension API now. Furthermore, I also removed the hand written ``get_document_width()`` methods (see commit message).

So, finally we now have:
https://github.com/jcwinkler/textext/blob/19ceb80194c2e1432fbd7cfd091d90771052101f/extension/textext/__init__.py#L395-L408

Short checklist:
- [x] Tested with Inkscape version: 1.0beta2 (Python 3.8)
- [ ] Tested on Linux, Distro: [fill in distribution name here]
- [x] Tested on Windows, Version: Windows 8.1

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -361,17 +361,13 @@ try:
                 if isinstance(text, bytes):
                     text = text.decode('utf-8')
 
-                # Coordinates in node from converter are always in pt, we have to scale them such that the node size is correct
-                # even if the document user units are not in pt
-                scale_factor = user_scale_factor * self.svg.unittouu("1pt")
-
                 # Convert
                 with logger.debug("Converting tex to svg"):
                     with ChangeToTemporaryDirectory():
                         converter = TexToPdfConverter(self.requirements_checker)
                         converter.tex_to_pdf(tex_executable, text, preamble_file)
                         converter.pdf_to_svg()
-                        tt_node = TexTextElement(converter.tmp("svg"))
+                        tt_node = TexTextElement(converter.tmp("svg"), self.svg.unittouu("1mm"))
 
                 # -- Store textext attributes
                 tt_node.set_meta("version", __version__)
@@ -398,10 +394,10 @@ try:
                         node_center = Vector2d(tt_node.get_center_position())
                         view_center = self.svg.get_center_position()
 
-                        tt_node.transform = (Transform(translate=view_center) *  # place at view center
-                                             Transform(scale=scale_factor) *  # scale
-                                             Transform(translate=-node_center) *  # place node at origin
-                                             tt_node.transform  # use original node transform
+                        tt_node.transform = (Transform(translate=view_center) *    # place at view center
+                                             Transform(scale=user_scale_factor) *  # scale
+                                             Transform(translate=-node_center) *   # place node at origin
+                                             tt_node.transform                     # use original node transform
                                              )
 
                         tt_node.set_meta('jacobian_sqrt', str(tt_node.get_jacobian_sqrt()))
@@ -427,9 +423,7 @@ try:
                     else:
                         self.config['preamble'] = ''
 
-                    # ToDo: Do we really need this if statement?
-                    if scale_factor is not None:
-                        self.config['scale'] = user_scale_factor
+                    self.config['scale'] = user_scale_factor
 
                     self.config["previous_tex_command"] = tex_command
 
@@ -608,11 +602,16 @@ try:
     class TexTextElement(inkex.elements.Group):
         tag_name = "g"
 
-        def __init__(self, svg_filename=None):
+        def __init__(self, svg_filename, uu_in_mm):
+            """
+            :param svg_filename: The name of the file containing the svg-snippet
+            :param uu_in_mm: The units of the document into which the node is going to be placed into
+                             expressed in mm
+            """
             super(TexTextElement, self).__init__()
-            self._svg_to_textext_node(svg_filename)
+            self._svg_to_textext_node(svg_filename, uu_in_mm)
 
-        def _svg_to_textext_node(self, svg_filename):
+        def _svg_to_textext_node(self, svg_filename, doc_unit_to_mm):
             from inkex.elements import ShapeElement, Defs
             from inkex.svg import SvgDocumentElement
             doc = etree.parse(svg_filename, parser=inkex.elements.SVG_PARSER)
@@ -628,6 +627,9 @@ try:
                 self.append(el)
 
             self.make_ids_unique()
+
+            # Ensure that snippet is correctly scaled according to the units of the document
+            self.transform.add_scale(doc_unit_to_mm/root.unittouu("1mm"))
 
         @staticmethod
         def _expand_defs(root):
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -194,32 +194,6 @@ try:
                 default=self.config.get('scale', 1.0)
             )
 
-        # Identical to inkex.Effect.getDocumentWidth() in Inkscape >= 0.91, but to provide compatibility with
-        # Inkscape 0.48 we implement it here explicitly again as long as we provide compatibility with that version
-        def get_document_width(self):
-            width = self.document.getroot().get('width')
-            if width:
-                return width
-            else:
-                viewbox = self.document.getroot().get('viewBox')
-                if viewbox:
-                    return viewbox.split()[2]
-                else:
-                    return '0'
-
-        # Identical to inkex.Effect.getDocumentHeight() in Inkscape >= 0.91, but to provide compatibility with
-        # Inkscape 0.48 we implement it here explicitly again as long as we provide compatibility with that version
-        def get_document_height(self):
-            height = self.document.getroot().get('height')
-            if height:
-                return height
-            else:
-                viewbox = self.document.getroot().get('viewBox')
-                if viewbox:
-                    return viewbox.split()[3]
-                else:
-                    return '0'
-
         def effect(self):
             """Perform the effect: create/modify TexText objects"""
             from asktext import AskerFactory
@@ -407,7 +381,6 @@ try:
                 tt_node.set_meta("preamble", preamble_file)
                 tt_node.set_meta("scale", str(user_scale_factor))
                 tt_node.set_meta("alignment", str(alignment))
-
                 try:
                     inkscape_version = self.document.getroot().get('inkscape:version')
                     tt_node.set_meta("inkscapeversion", inkscape_version.split(' ')[0])
@@ -416,27 +389,28 @@ try:
                     # no version attribute is provided by Inkscape :-(
                     pass
 
-                # -- Copy style
+                # Place new node in document
                 if old_svg_ele is None:
                     with logger.debug("Adding new node to document"):
-                        root = self.document.getroot()
-                        width = self.svg.unittouu(self.get_document_width())
-                        height = self.svg.unittouu(self.get_document_height())
-
+                        # Place new nodes in the center of the document and scale them according to user request
+                        doc_width = self.svg.width
+                        doc_height = self.svg.height
                         x, y, w, h = tt_node.bounding_box()
                         tt_node.transform = tt_node.transform * \
-                                            Transform(translate=(-x + width / 2 - w / 2, -y + height / 2 - h / 2)) * \
+                                            Transform(translate=(-x + doc_width / 2 - w / 2,
+                                                                 -y + doc_height / 2 - h / 2)) * \
                                             Transform(scale=scale_factor)
                         tt_node.set_meta('jacobian_sqrt', str(tt_node.get_jacobian_sqrt()))
 
                         self.svg.get_current_layer().add(tt_node)
                 else:
                     with logger.debug("Replacing node in document"):
+                        # Rescale existing nodes according to user request
                         relative_scale = user_scale_factor / original_scale
                         tt_node.align_to_node(old_svg_ele, alignment, relative_scale)
 
-                        # If no non-black color has been explicitily set by TeX we copy the color information from the old node
-                        # so that coloring done in Inkscape is preserved.
+                        # If no non-black color has been explicitily set by TeX we copy the color information
+                        # from the old node so that coloring done in Inkscape is preserved.
                         if not tt_node.is_colorized():
                             tt_node.import_group_color_style(old_svg_ele)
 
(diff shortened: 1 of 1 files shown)


extension/textext/__init__.py
```diff
@@ -416,27 +389,35 @@ def do_convert(self, text, preamble_file, user_scale_factor, old_svg_ele, alignm
...
+                        node_cx, node_cy = tt_node.get_center_position()
+                        node_cx = node_cx*self.svg.unittouu("1pt")
+                        node_cy = node_cy*self.svg.unittouu("1pt")
 
-                        x, y, w, h = tt_node.bounding_box()
                         tt_node.transform = tt_node.transform * \
-                                            Transform(translate=(-x + width / 2 - w / 2, -y + height / 2 - h / 2)) * \
+                                            Transform(translate=(doc_cx - node_cx, doc_cy - node_cy)) * \
                                             Transform(scale=scale_factor)
+
```
sizmailov: I tested it with number of document units and scales and it doesn't seem to do right thing. 

1) `node.get_center_position()` returns in user units already
2) `svg.get_center_position()` returns *view* center, not document (bad naming in `inkex`, I'll file an issue)
3) scale was applied *before* compensation of the node width

The following snippet does the trick. Note: now node is placed in middle of view, not in center of document as before. I think it's even better

```python
                        # Place new nodes in the view center and scale them according to user request

                        from inkex.transforms import Vector2d

                        node_center = Vector2d(tt_node.get_center_position())
                        view_center = self.svg.get_center_position()

                        tt_node.transform = (Transform(translate=view_center) *   # place at view center
                                             Transform(scale=scale_factor) *      # scale 
                                             Transform(translate=-node_center) *  # place node at origin
                                             tt_node.transform                    # use original node transform
                                             )
```

jcwinkler: Thanks for the careful review. My mistake was to test only with a fully visible A4 page centered in the screen after pressing "5" so center of view is equal to center of page. And scaling before moving, you are right, of course.

However, I cannot reproduce your statement

>     1. `node.get_center_position()` returns in user units already

Under both, Linux and Windows the node data is returned in pt in by environment.

Furthermore, despite what I am setting under Inkscape - Document properties, user units are always in mm. Under Edit - Preferences I do not find anything. What I am missing, where is my misunderstanding?

jcwinkler: > Note: now node is placed in middle of view, not in center of document as before. I think it's even better

This is definitely a great enhancement!

sizmailov: > 1. `node.get_center_position()` returns in user units already

I cite here doc string of the function. I tried different units in document properties. The size of output was same and matched text objects created by inkscape. 

Do you observe any misbehavior? As long as we use one and the same units it doesn't matter which one.

sizmailov: > This is definitely a great enhancement!

It should be reflected in docs and change list. Not necessarily as part of this PR, but it will be better.

jcwinkler: Doing an ``etree.dump(root)`` after import of the generated node svg at

https://github.com/textext/textext/blob/ccf023ed75b77a210c6d005f8998df881001c306/extension/textext/__init__.py#L637-L642

one obtains (compiled LaTeX snippet: ``$\times$``)

```xml
<svg xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape" sodipodi:docname="tmp.pdf" id="svg17" version="1.2" viewBox="0 0 4.8125 4.796875" height="4.796875pt" width="4.8125pt">
  <metadata id="metadata21">
    <rdf:RDF>
      <cc:Work rdf:about="">
        <dc:format>image/svg+xml</dc:format>
        <dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage"/>
      </cc:Work>
    </rdf:RDF>
  </metadata>
  <sodipodi:namedview id="namedview19" inkscape:window-height="480" inkscape:window-width="640" inkscape:pageshadow="2" inkscape:pageopacity="0" guidetolerance="10" gridtolerance="10" objecttolerance="10" borderopacity="1" inkscape:document-rotation="0" bordercolor="#666666" pagecolor="#ffffff"/>
  <defs id="defs10">
    <g id="g8">
      <symbol id="glyph0-0" overflow="visible">
        <path inkscape:connector-curvature="0" id="path2" d="" style="stroke:none"/>
      </symbol>
      <symbol id="glyph0-1" overflow="visible">
        <path inkscape:connector-curvature="0" id="path5" d="M 3.875,-2.765625 1.890625,-4.75 C 1.765625,-4.875 1.75,-4.890625 1.671875,-4.890625 1.5625,-4.890625 1.46875,-4.8125 1.46875,-4.6875 c 0,0.0625 0.015625,0.078125 0.125,0.1875 l 2,2 -2,2.015625 c -0.109375,0.109375 -0.125,0.125 -0.125,0.1875 0,0.125 0.09375,0.203125 0.203125,0.203125 0.078125,0 0.09375,-0.015625 0.21875,-0.140625 L 3.875,-2.21875 l 2.0625,2.0625 c 0.015625,0.015625 0.078125,0.0625 0.140625,0.0625 0.125,0 0.203125,-0.078125 0.203125,-0.203125 0,-0.015625 0,-0.046875 -0.03125,-0.109375 C 6.234375,-0.421875 4.65625,-1.984375 4.15625,-2.5 l 1.828125,-1.8125 c 0.046875,-0.0625 0.203125,-0.1875 0.25,-0.25 0,-0.015625 0.046875,-0.0625 0.046875,-0.125 0,-0.125 -0.078125,-0.203125 -0.203125,-0.203125 C 6,-4.890625 5.96875,-4.859375 5.859375,-4.75 Z m 0,0" style="stroke:none"/>
      </symbol>
    </g>
  </defs>
  <g transform="translate(-150.18076,-129.87437)" id="surface1">
    <g id="g14" style="fill:#000000;fill-opacity:1">
      <use height="100%" width="100%" id="use12" y="134.765" x="148.71201" xlink:href="#glyph0-1"/>
    </g>
  </g>
</svg>
```

Clearly, the units are interpreted as pt here and from the viewpoint of the API these are the user units of the snippet (The command line conversion pdf->svg always results in snippets the user units of which are in pt as it seems.) But they are not related to the document you are currently working with. 

What I do not understand is:
- ``tt_node.get_center_position()`` returns the position in pt
- ``elf.svg.get_center_position()`` returns the view center in, for example, mm

but when consecutively apply the transform we can mix the translate operations (one in pt, one mm).

Here is the situation in a debugging screenshot of the text ``Hello world!`, scale factor 1.0`:

grafik

You can reproduce it in a document of A4-size, fully centered in the screen (cx = 105mm, cy 148mm). The dimension of the snippet is w = 18.41mm, h=2.55mm or **in pt: w = 51.24pt, h = 7.25pt** the half of which matches the value in the screenshot of the debugging session.

Do you have any clue? (Obviously, this works, as we see, but why, that's my problem... :confused: )

jcwinkler: OK, I can give the answer for myself:

Node:
```bash
>> tt_node.transform
Transform(((1, 0, 0), (0, 1, 0)))
```` 

Move operation in pt:
```bash
>>Transform(translate=-node_center)*tt_node.transform
Transform(((1, 0, -25.7122), (0, 1, -3.62537)))
```

By applying the scalefactor the move operation is in fact in mm (in this example):
```bash
>>Transform(scale=scale_factor)*Transform(translate=-node_center)*tt_node.transform
Transform(((0.352778, 0, -9.07069), (0, 0.352778, -1.27895)))
```

And then the movment to the document center in mm
```bash
>>Transform(translate=view_center)*Transform(scale=scale_factor)*Transform(translate=-node_center)*tt_node.transform
Transform(((0.352778, 0, 96.7626), (0, 0.352778, 146.888)))
```

So this makes really sense!

In order to not rely on a hard-coded unit for the snippet maybe we should get the unit by
```python
# height="7.24pt" -> 'pt'
inkex.units.parse_unit(root.get('height'), 'pt', 1)[1]
```
directly after importing the snippet? Will fallback to 'pt' if no height attrib is found. What do you think?

sizmailov: Ok. Now I see where 1pt came from! The right thing to do is to scale node at the import time to factor out snippet units. To find out ratio we can convert anything to user units of snippet and to user units of root document: 

```python
# pseudo-code
snippet_to_root_ratio = snippet.unittouu("1mm") / self.svg.unittouu("1mm")
node.transform.add_transform(1/snippet_to_root_ratio) 
```

It should be done in `TexTextElement._svg_to_textext_node()`. Root document will become constructor argument

jcwinkler: OK, I can give the answer for myself:

Node:
```bash
>> tt_node.transform
Transform(((1, 0, 0), (0, 1, 0)))
```` 

Move operation in pt:
```bash
>>Transform(translate=-node_center)*tt_node.transform
Transform(((1, 0, -25.7122), (0, 1, -3.62537)))
```

By applying the scalefactor the move operation is in fact in mm (in this example):
```bash
>>Transform(scale=scale_factor)*Transform(translate=-node_center)*tt_node.transform
Transform(((0.352778, 0, -9.07069), (0, 0.352778, -1.27895)))
```

And then the movment to the document center in mm
```bash
>>Transform(translate=view_center)*Transform(scale=scale_factor)*Transform(translate=-node_center)*tt_node.transform
Transform(((0.352778, 0, 96.7626), (0, 0.352778, 146.888)))
```

So this makes really sense!

In order to not rely on a hard-coded unit for the snippet maybe we should get the unit by
```python
# height="7.24pt" -> 'pt'
inkex.units.parse_unit(root.get('height'), 'pt', 1)[1]
```
directly after importing the snippet? Will fallback to 'pt' if no height attrib is found. What do you think?

jcwinkler: We were on the same track...

> Ok. Now I see where 1pt came from! The right thing to do is to scale node at the import time to factor out snippet units. To find out ratio we can convert anything to user units of snippet and to user units of root document

Yeah, that's the most elegant way. Now it is factored out in the consecutive transform operations. I will provide an implementation.

jcwinkler: This makes the scaling of the scale factor unnecessary, too. Things become consistent now...

sizmailov: Yeah, units. I remember I had doubts about my modelling program and I spent days looking for origins of 2 pi factor... :crying_cat_face:

sizmailov: FYI: I [proposed](https://gitlab.com/inkscape/extensions/merge_requests/157) removal of `.get_center_position()` methods from inkex API

The updated version should look like
```python
node_center = tt_node.bounding_box().center # works now
view_center = self.svg.namedview.center # not yet available 
```

sizmailov: @jcwinkler All proposed changes are in master branch of `inkscape/extensions`. I don't know when they show up in appimages. May be you should delay changes until it happen. 

We may need finer control of inkscape version, because the change will be not compatible with alpha versions of inkscape.

jcwinkler: As far as I do understand the beta 2 Inkscape versions for Windows are continously updated from the master branch, the Linux AppImages not: https://inkscape.org/release/inkscape-1.0beta2/ .

sizmailov: Center calculation needs to be changed to
```python
node_center = tt_node.bounding_box().center
view_center = self.svg.namedview.center
```

extension/textext/__init__.py
```diff
@@ -608,11 +602,16 @@ def parse_pdf_log(self, logfile):
     class TexTextElement(inkex.elements.Group):
         tag_name = "g"
 
-        def __init__(self, svg_filename=None):
+        def __init__(self, svg_filename=None, doc_unit_to_mm=1.0):
```
sizmailov: 1) `doc_unit_to_mm` is actually `1mm` in document units, should be named accordingly, e.g. `mm_in_uu` or `mm_in_doc_units`
2) Why `doc_unit_to_mm` has default value?

extension/textext/__init__.py
```diff
@@ -651,6 +628,9 @@ def _svg_to_textext_node(self, svg_filename):
 
             self.make_ids_unique()
 
+            # Ensure that snippet is correctly scaled according to the units of the document
+            self.transform.add_scale(doc_unit_to_mm/root.unittouu("1mm"))
```
sizmailov: We want to scale by ratio of doc units to root units. It seems like we should scale by inverse of what written.
```python 
doc/root = (doc/mm) / (root/mm) = ( mm / root ) / ( mm / doc) = mm_to_root / mm_to_doc
```

So it should be
```
self.transform.add_scale(root.unittouu("1mm")/mm_to_doc)
```
, right?

jcwinkler: No, the other way round as in the commit. ``root.unittouu("1mm")/mm_to_doc`` leads to completely wrong scaling (much too large).

jcwinkler:
I implement this workaround so we can use the extension in both, official beta versions and development code of Inkscape. ToDo remark added in Project: https://github.com/textext/textext/projects/1#card-31075018

extension/textext/__init__.py
sizmailov: Found where I was mistaken. `svg.unittouu("1mm")` returns "user_units_per_mm", not 
"mm_per_user_unit".

sizmailov:
Found another problem. 
1. Rotate canvas (Ctrl+Shift+mouse wheel) 
2. Run textext

New node placed far away from away from view point center.

sizmailov:
Tested on 1.1-dev (3322b9cb8d, 2019-12-31)

extension/textext/__init__.py
sizmailov: I suggest to rename `mm_to_doc` to less confusing `user_units_per_mm`.

sizmailov:
Rotation issue is inkscape bug. I submitted a MR to fix

PR #171, opened 2019-12-20
Fix Ctrl+P shortcut

sizmailov:
Preview was triggered by bare `P`, not `Ctrl+P`

Short checklist:
- [X] Tested with Inkscape version: [1.1-dev (d3cbf63a92, 2019-12-19)]
- [X] Tested on Linux, Distro: [Ubuntu 18.04]


jcwinkler:
Don't we have the same problem with CTRL+Q?
https://github.com/textext/textext/blob/ccf023ed75b77a210c6d005f8998df881001c306/extension/textext/asktext.py#L729-L731

jcwinkler:
My observation without your fix: Under Python 3.8 it works, under Python 2.7 it does not work (even with bare p)

sizmailov:
> Don't we have the same problem with CTRL+Q?

Seems so

jcwinkler:
Could you update your PR accordingly, please. Then I merge it.

sizmailov:
Tested with python2 + python3

Issue #161, opened 2019-12-05
Refactor of method names in asktext.py


jcwinkler:
Several method names in the GUI classes are not intuitive/ meaningful. This should be refactored.

Example:
https://github.com/textext/textext/blob/1486c7febd9d142693c2b4cc6bc5022a3b927305/extension/textext/asktext.py#L707

Meant is that the shortcut is changed.

https://github.com/textext/textext/blob/1486c7febd9d142693c2b4cc6bc5022a3b927305/extension/textext/asktext.py#L736

Meant is that the OK button is clicked.

etc., etc.

PR #176, opened 2020-01-02
Refactor/latex error handling

jcwinkler:
I observed that latex compiling errors lead to an empty error message window. Reason is that the latexlogparser does not write the parsed error message into the logstream since it is disabled by default. Hence I changed the disable level temporarily (it set to to logging.CRITICAL if the extensions runs normally):

https://github.com/textext/textext/blob/c3ac8d9da8a4153f208edce4ed4769f169ad5c71/extension/textext/__init__.py#L596-L600

I am not really satisfied with this solution. Do have a better idea, @sizmailov ?

Furthermore, when the extension runs under Python 2.7 a unicode error occurs when the error message is formatted. Hence I force a unicode string:

https://github.com/textext/textext/blob/551698c7a2ee2ff8b50e5b3112d210d7b640ee17/extension/textext/typesetter.py#L103-L105

Again, not very elegant (what, if for some reason the stream wants to have a standard string?). Any suggestions?

Maybe we should drop the typesetter/latexlogparser approach and use the much more simple library https://github.com/inakleinbottle/texoutparse ? In fact we are only interested in the message string. (Personally, I would prefer this way)

Other topics going to be covered by this PR:
- stdout box in error dialog not scrollable (problem when long output is shown)
- Correction of line-number offset in error messages


jcwinkler:
(Hmm texoutparse is Python 3 only :angry:)

sizmailov:
Latex error handling now is not good. Error window is unscrollable, errors are not shown, etc. 

I think it's better to refactor/replace `typesettrer` & `latexlogparser`, rather than patching textext. In principle we can fall back to plain output for python2 if needed...

jcwinkler:
OK. I will remove ``typesetter`` and ``latexlogparser`` and provide a smarter solution.

sizmailov:
What about forking `textoutparse` to `textext/textoutparse` and adding fork as submodule?

extension/textext/asktext.py
```diff
@@ -483,11 +427,62 @@ def select_preamble_file(self):
...
+
+            # ToDo: Check Windows behavior!! --> -disable
+            self._root.wm_attributes("-topmost", False)
+
+            self._err_dialog = Tk.Toplevel(self._frame)
+            self._err_dialog.minsize(300, 400)
+            self._err_dialog.transient(self._frame)
+            self._err_dialog.focus_force()
+            self._err_dialog.grab_set()
+            self._err_dialog.protocol("WM_DELETE_WINDOW", error_dialog_close)
```
sizmailov: Here `error_dialog_close` can capture local `err_dialog` just being moved down by 30 lines. (also `close_error_dialog` is better name)

Try to not store temporary variables as `self` hidden `attributes` whenever possible. It will reduce number of "states" of `self` and make reasoning about program easier.

extension/textext/asktext.py
```diff
@@ -483,11 +427,62 @@ def select_preamble_file(self):
...
+
+            add_textview(label, str(error))
+
+            if isinstance(error, (TexTextConversionError, TexTextCommandFailed)):
```
sizmailov: `TexTextConversionError` is subclass of `TexTextCommandFailed`, remove former

extension/textext/asktext.py
```diff
@@ -758,10 +753,9 @@ def cb_ok(self, widget=None, data=None):
             except Exception as error:
                 import traceback
```
sizmailov: unused import

extension/textext/asktext.py
```diff
@@ -483,11 +427,62 @@ def select_preamble_file(self):
...
+        def error_dialog(self, title, label, error):
+            """
+            Present an error dialog
+
+            :param parent: Parent window
+            :param title: Error title text
+            :param label: Label text
+            :param error: exception
+            :type error: StandardError
+            """
```
sizmailov: Don't repeat superclass's docstring

extension/textext/asktext.py
```diff
@@ -1247,3 +1232,79 @@ def ask(self, callback, preview_callback=None):
...
+                expander.set_label(header)
+                expander.set_use_markup(True)
+
+                expander.set_size_request(20, -1)
+                scroll_window.hide()
+
+                dialog.vbox.pack_start(expander, expand=True, fill=True, padding=5)
+
+            dialog.vbox.pack_start(message_label, expand=False, fill=True, padding=5)
+            message_label.show()
```
sizmailov: Is this line needed?

extension/textext/asktext.py
```diff
@@ -1247,3 +1232,79 @@ def ask(self, callback, preview_callback=None):
...
+        def error_dialog(self, title, label, error):
+            """
+            Present an error dialog
+
+            :param parent: Parent window
+            :param title: Error title text
+            :param label: Label text
+            :param error: exception
+            :type error: StandardError
+            """
```
sizmailov: Don't repeat superclass's docstring

sizmailov:
I know it's still work in progress. Just added some generic comments, didn't test it so far.

jcwinkler:
Thanks for the hints! The code is full of ambiguous naming, inconsistent docstrings and unused imports. Hence, I will first merge this PR after having addressed your comments together with #177 into develop. Then I will focus on all this stuff (see also #161) in a separate draft PR so you can closely follow and we achieve an acceptable result. I already started in https://github.com/jcwinkler/textext/tree/refactor/cleanup_2019 (needs to be rebased, first, of course).

extension/textext/asktext.py
jcwinkler: Right. Also applies to GTK-version.

extension/textext/asktext.py
```diff
@@ -246,6 +169,18 @@ def ask(self, callback, preview_callback=None):
...
 
+    def error_dialog(self, title, label, error):
+        """
+        Presents an error dialog
+
+        :param parent: Parent window
+        :param title: Error title text
+        :param label: Label text
+        :param error: exception
+        :type error: StandardError
```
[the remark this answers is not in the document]
jcwinkler: Wouldn't the consequent usage of``abstractmethod``be more consistent?

sizmailov:
ok, cool. Let me know when it will be ready for review.

extension/textext/asktext.py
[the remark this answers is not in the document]
sizmailov: I think it's not worth here. The base class and Tk live their last days before we switch to GTK3 completely. 

In principle `abc` provides means to express class hierarchy in more details than bare `raise NotImplementedError`. Here we are authors and the only consumers of the class, no need to overdocument it. 

I don't like `pass` because it says "you are free to implement this method, but you not have to"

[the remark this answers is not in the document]
jcwinkler: :heavy_check_mark:

extension/textext/asktext.py
jcwinkler: No. 

:heavy_check_mark:

extension/textext/asktext.py
jcwinkler: :heavy_check_mark:

extension/textext/asktext.py
jcwinkler: :heavy_check_mark:

extension/textext/asktext.py
jcwinkler: :heavy_check_mark:

jcwinkler:
I will merge this together with #177 into develop and then do final cleanup in  https://github.com/jcwinkler/textext/tree/refactor/cleanup_2019

sizmailov:
Do you need help with using `texoutparse` as submodule?

jcwinkler:
> Do you need help with using `texoutparse` as submodule?

No, thank you. I know the handling of submodules. However, I had to tweak the code manually such that it is also compatible with Python2. Hence, just including the submodule does not help.

Two suggestions:

1. We include the Python3 version of texoutparse as submodule and also provide a tweaked version in our repo, e.g. as texoutparse2.

2. I contact the developer to provide such a version in his repo.

What do you think?

sizmailov:
1. Fork `texoutparse` to `https://github.com/textext/texoutparse`
2. Create branch `adjust-to-python-2` and commit your changes
3. Add as a submodule to main textext repository and ping to your branch 
4. (optional) submit PR to main `texoutparse` repo

jcwinkler:
OK, that's another good possibility.

sizmailov:
Did you merge by accident?

jcwinkler:
No. I will do the submodule approach separately.

Issue #173, opened 2019-12-29
Tk status in TexText 1.0 


sizmailov:
Tk is guaranteed to be a part of every python interpreter, so TexText may run even if GTK libraries are missing. At the same time the functionality and interface are very limited compared to GTK. Now (inkscape>=1.0) guarantees that GTK3 is installed, so I think it might be a good point to deprecate Tk or remove entirely. 

Is there any user which still **need** Tk support in inkcape1.0+? (please leave a comment if you need tk and tell us what is your use case)

sizmailov:
@jcwinkler You were against TK deprecation in TexText 1.0 as GTK was shaky in inkscape 1.0. Is current situation still same? 

I assume GTK+PyGObject are reliably available on all platforms as inkscape 1.0 release appoaches.
Inkscape's release date was shifted number of times, but now it seems to be really about release which is around 4th of May.

As a compromise we can keep Tk version in 1.x in unsupported mode and plan removal to 2.x.
In 1.0 we can place deprecation warning in Tk UI to get feedback about it, smth like:

```
TexText does not longer support Tk UI.  
Please see deprecation page notice on textext.github.com/textext/tk-deprecation
```

On deprecation page we can provide a link to this issue to and ask people who need Tk describe their use case.

jcwinkler:
I agree that we should drop Tk support as soon as possible. However, to be on the safe side let's keep it in the 1.x branch using the approach you sketched and without developing it further.

I will add this to the doc I am also currently updating.

On Windows it is guaranteed that GTK+pyGObject is included. Currently I am building an installer providing also GTKSourceview.

On Linux the situation is quite unclear to me:

- Will gi retroinspection etc. be included as a dependency in the deb/rpm packages? Or do the user still need to install it?
- Will gi retroinspection etc. be included in the Appimages, flatpack and snap packages?

Issue #150, opened 2019-11-10
Textext (develop branch) not working with Inkscape 1.0 beta on MAC running Catalina


yoramzarai:
SW versions:
Inkscape: 1.0beta1 version
MAC OS: Catalina (10.15.1)
Textext - develop branch version 

Description:
1. In Inkscape, upon clicking Extensions -> Tex Text, it complains that pdflatex not in path ("`pdflatex` is NOT found in PATH"). However, pdflatex is in path ("which pdflatex" in the terminal yields "/opt/local/bin/pdflatex"). 
2. I then changed "path = ["/Applications/Inkscape.app/Contents/Resources"]" in requirements_check.py to "path = ["/Applications/Inkscape.app/Contents/Resources","/opt/local/bin"] ", relaunched Inkscape, and now upon Extensions -> Tex Text, the "TexText 0.11.0" windows opens and when entering a Latex code at the appropriate place (e.g. $\alpha$) and clicking OK, I get a window with the following :
2019-11-10 15:14:48.767 Python[36004:1024964] CoreText note: Client requested name ".SFNSMono-Regular", it will get Times-Roman rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2019-11-10 15:14:48.767 Python[36004:1024964] CoreText note: Set a breakpoint on CTFontLogSystemFontNameRequest to debug.
2019-11-10 15:14:48.768 Python[36004:1024964] CoreText note: Client requested name ".SF NS Mono", it will get Times-Roman rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2019-11-10 15:14:48.851 Python[36004:1024964] CoreText note: Client requested name ".SFNSMono-Regular", it will get Times-Roman rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2019-11-10 15:14:48.851 Python[36004:1024964] CoreText note: Client requested name ".SF NS Mono", it will get Times-Roman rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
[...]

PR #177, opened 2020-01-02
Enabled TkInter under Python 3

jcwinkler:
Related issue(s): #173 

This PR re-enables TkInter support when the extension is run under Python 3. It is good to have this as backup since installation of gtk-bindings is still problematic (see e.g. discussion in #150)

Short checklist:
- [x] Tested with Inkscape version: 1.0 beta 2
- [x] Tested on Linux Kubuntu 18.04 (Python 3.6)

### DIFF
--- a/extension/textext/asktext.py
+++ b/extension/textext/asktext.py
@@ -41,7 +41,7 @@ from textext.utility import SuppressStream
 
 # Try GTK first
 #   If successful, try GTKSourceView (bonus points!)
-#   If unsuccessful, try TK
+#   If unsuccessful, try TK (first for Python 3, then for Python 2)
 #   When not even TK could be imported, abort with error message
 try:
     import gi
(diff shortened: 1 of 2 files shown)


sizmailov:
Currently non-functional GTK3 is inkscape bug, and its going to be fixed. From this moment I see no reason to support less-featured 90x-looking interface.

Tk library is provided by python, therefore guaranteed to work everywhere. Now when GTK3 is provided by inkscape 1.1+ and TK (almost) lost availability advantage, I don't see any reasons to keep it.

If you still want to go this route, note that `requirements_check` also requires correction

jcwinkler:
As long as GTK GUI is not proven to work out of the box in the wildness (unexperienced, impassient users) I will keep the TK stuff. It is not too much work. Anyway, I 100% agree that it is old styled and should removed as soon as possible.

OT: Is there a document explaining what GTK3 support means? 
- Python3 GTK bindings are included in Windows version.
- Will it be an install dependency under Linux?
- What is with GTK sourceview? Under Windows I have to battle that they include it in the package, otherwise I need to provide an installer.

sizmailov:
>  OT: Is there a document explaining what GTK3 support means?

I don't believe it exists. The best way to get correct answer is to ask in dev chat. 

My current understanding of GTK availability:
 - inkscape depends on GTK3, so all shared libraries, etc are guaranteed to be there and work for 1.0+
 - `PyGObject`  is guaranteed to be available  starting from 1.1

>    What is with GTK sourceview?

I think it's not going to be included as it's not needed by inkscape. Therefore it must be installed separately.

extension/textext/asktext.py
```diff
@@ -62,17 +62,24 @@
 
 except (ImportError, TypeError, ValueError) as _:
     try:
-        import Tkinter as Tk
```
sizmailov: Can we branch on `sys.version_info[0]` instead ?

extension/textext/requirements_check.py
```diff
@@ -548,11 +548,17 @@ def find_pygtk3(self):
...
 
     def find_tkinter(self):
+        executable = sys.executable
         try:
-            executable = sys.executable
-            defaults.call_command([executable, "-c", "import Tkinter; import tkMessageBox; import tkFileDialog;"])
+            # Python 3
+            defaults.call_command(
+                [executable, "-c", "import tkinter; import tkinter.messagebox; import tkinter.filedialog;"])
         except (KeyError, OSError, subprocess.CalledProcessError):
```
sizmailov: Why is `KeyError` is here?

sizmailov: I would replace nested try catch with smth like this:
```python
if sys.version_info[0]==2:
    import_tk_script = "..."
else: # python3
    import_tk_script = "..."

PR #178, opened 2020-01-09
Refactor/cleanup 2019

jcwinkler:
This draft pull request is intended to enable everyone to follow some refactoring work and to give constructive comments/ hints if needed.

Currently, the following topics are going to be addressed:

- [x] Remove of unused imports
- [x] Remove of unneeded global variables
- [ ] Check method and variable names if they are meaningful (really express what is done/ represented)
- [ ] Cleanup of docstrings
- [ ] Consistent file headers (remove old comments, etc.)
- [ ] Check for unneeded class members (replace by locals if possible)
- [ ] Other cosmetic stuff like wrong indents, double/ missing whitespaces, etc (pylint is quite verbose at the moment :smile: ).
- [ ] create ``textext/__main__.py`` and move run logic there, leaving ``__init__.py`` with bare class definitions

t.b.c.

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -38,8 +38,6 @@ florent becker and vladislav gavryusev for contributions.
 .. _inkscape: http://www.inkscape.org/
 .. _inklatex: http://www.kono.cis.iwate-u.ac.jp/~arakit/inkscape/inklatex.html
 """
-
-from __future__ import print_function
 import hashlib
 import logging
 import logging.handlers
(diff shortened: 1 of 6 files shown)


jcwinkler:
@sizmailov Do we really need this assignment?

https://github.com/textext/textext/blob/94b49b768eefa50161183defef9c5fa90b1ea06e/extension/textext/__init__.py#L65-L66

I do not see any place where we instanciate an unicode object and I do not observe any issues when running the extension under Python3.

sizmailov:
Yes, looks like it's not used anymore.

sizmailov:
Add to the list:
 - create `textext/__main__.py` and move run logic there, leaving `__init__.py` with bare class definitions

jcwinkler:
I know much of the stuff committed so far is more or less a matter of taste but I feel much better with this cleanup now.

In fact, TexTextLogger is also usable in other contexts so maybe the name is too specific. But better  I prefer this compared to MyLogger.

PR #184, opened 2020-02-10
Added GUI binding installation instructions for beta releases

jcwinkler:
@sizmailov I just extended the installation instructions for TexText development versions by the GTK bindings. What do you think?

Please see https://jcwinkler.github.io/textext/install/linux-beta.html and https://jcwinkler.github.io/textext/install/windows-beta.html

Note that for the final release (i.e. without the AppImage) I provide a completely rewritten version of the help including new screenshots, troubleshooting section etc. I will inform you as soon as it is ready for review.

### DIFF
--- a/docs/source/install/linux-beta.rst
+++ b/docs/source/install/linux-beta.rst
@@ -73,6 +73,67 @@ executing
     ~/squashfs-root/bin/inkscape &
 
 
+Install GUI library
+===================
+
+Install the Python bindings for the graphical user interface of
+|TexText|. You have two options: ``GTK3`` (recommended) or ``Tkinter``.
+
+At first you need to discover the Python interpreter that is used by your
+Inkscape installation. Enter the following command in a terminal
+
+.. code-block:: bash
+
+        python --version
+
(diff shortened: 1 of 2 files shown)


sizmailov:
Hi, I didn't read through it thoroughly here is just a couple of things (not related to GUI update in current PR):

- Preparation can be simplified by using `mv` command 
- It's better to avoid pollution of root directory (`~/squashfs-root`). I would consider moving it to `./local/share/inkscape-beta/` and provided a symlink in `~/bin/` or `./local/bin/`. Then beta version of inkscape will be discoverable (and become default) without modifying PATH on most systems (of course after re-login). It would also simplify "download" section.
 - I would add a small section with instruction how to run older version of inkscape if needed. 
```bash
# discover current full path to inkscape to be ease access to older version if needed
which inkscape
```

Didn't check packages names, but looks correct. 

As all my comments are irrelevant to PR I think it's ok to merge

jcwinkler:
Ok thanks. Good points. I will add them to the improved doc.

Issue #188, opened 2020-02-23
TexText 1.0: Unknown options --without-gui, --export-file


orengt:
Just installed TexText-Windows-1.0.0-dev.2 for: inkscape-1.0beta2_2020-02-06_235953b-x64.

I'm trying to create a formula.
I entered a simple one as $V_0$ and when preview/save I get the following error:

Command D:\Software Download\Inkscape\inkscape-1.0beta2_2020-02-06_235953b-x64\inkscape\bin\inkscape.exe --without-gui --pdf-poppler --pdf-page=1 --export-type=png --export-area-drawing --export-dpi=300 --export-file tmp.png tmp.pdf failed (code 1)

(org.inkscape.Inkscape:5472): GLib-GIO-WARNING **: 16:17:01.192: win32 session dbus binary not found: D:\Software Download\Inkscape\inkscape-1.0beta2_2020-02-06_235953b-x64\inkscape\bin\\gdbus.exe
Unknown option --without-gui

sizmailov:
Hi, thanks for report. inkscape recently removed `--without-gui` command line option. 

The fix would be to remove `--without-gui` option from `textext/__init__.py` (two occurrences)

Alternatively you can use inkscape beta snapshot before approximately mid-January (I believe).

orengt:
Sergei,

Thanks a lot for your quick response, it helps a lot.
I removed the options and it works.
Oren

Masacroso:
You must also change in __init__.py the string --export-file by --export-filename (it appear twice also). Then, with these two changes, the extension work for the actual beta2 version of inkscape.

jcwinkler:
Confirmed in most recent beta2 version. I'll fix it.

jcwinkler:
Unfortunately, these changes in the Inkscape extension API do only float into the Windows build of Inkscape beta2. This is built on a nightly basis whereas the Linux AppImage build reflects the state of December 3rd 2019 of Inkscape beta 2, hence does not have the change --export-file -> --export-filename.

(I will update the TexText 1.0 Windows package manually)

PR #192, opened 2020-03-03
Fixed: Unknown option --export-file

jcwinkler:
Inkscape command line option --export-filename has been changed to --export-filename in Inkscape.

Related issue(s): #188

This change is currently not available in Linux beta builds, only on Windows. Hence, I will keep this PR open until the option is available there.

Short checklist:
- [x] Tested with Inkscape version: 1.0 beta2
- [ ] Tested on Linux -> Test fails on current beta 2 from Dec 3 2019
- [x] Tested on Windows 8.1
- [ ] Tested on MacOS

### DIFF
--- a/extension/textext/__init__.py
+++ b/extension/textext/__init__.py
@@ -548,7 +548,7 @@ try:
                 "--export-type=svg",
                 "--export-text-to-path",
                 "--export-area-drawing",
-                "--export-file", self.tmp('svg'),
+                "--export-filename", self.tmp('svg'),
                 self.tmp('pdf')
             ]
             )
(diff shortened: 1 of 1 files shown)


sizmailov:
There are even more changes: `--extension-directory` is also gone now (superseded by `--system-data-directory`).  

Note we can guess perfectly where to put extension using `--user-data-directory` option. 

I think it's ok to be compatible with latest builds available in `1.0.x` branch. It would guarantee no divergence between beta versions provided for different OS. 

AppImage & DMG packages are build as part of gitlab CI:
https://gitlab.com/inkscape/inkscape/-/commits/1.0.x
It would require some clicks, but it's there

Windows builds:
https://ci.appveyor.com/project/inkscape/inkscape/history

<details>
<summary>clicks explained (AppImage) </summary>

image

image

image

image


</details>

jcwinkler:
OK, thanks @sizmailov. I will update the documentation so users can acess the most recent beta versins for all platforms.

sizmailov:
TBH I don't see reason to keep it open. It's dev branch after all and it works with most recent inkscape builds.

PR #236, opened 2020-07-17
Doc: Inlcude darkmode and inkscape 1.0 install instructions

jcwinkler:
@sizmailov I updated the doc for the next release:

- White background in darkmode themes ([see here](https://jcwinkler.github.io/textext/usage/gui.html#preview-button))

- Instructions how to install Inkscape 1.0 in order to help the average user ([see here](https://jcwinkler.github.io/textext/install/linux.html))


sizmailov:
The dark theme alignment icons should have light-gray "TeX" instead of solid black (pure white would be too bright and distracting).
Other than that looks good to me.

PR #248, opened 2020-08-21
Add "stroke-width=0pt" to SVG elements with "stroke=none"

jcwinkler:
For each element which has the style attribute "stroke"
set to "none" a style attribute "stroke-width" with
value "0" is added. This ensures that when colorizing
the node later in inkscape by setting the node and
stroke colors letters do not become bold (letters have
"stroke" set to "none" but e.g. horizontal lines in
fraction bars and square roots are only affected by
stroke colors so for full colorization of a node you
need to set the fill as well as the stroke color!).

**Colorized node (fill + stroke color) before this commit:**
grafik

**Colorized node (fill + stroke color) after this commit:**
grafik

What do you think @sizmailov?

Short checklist:
- [x] Tested with Inkscape version: 1.0
- [x] Tested on Linux Kubuntu 18.04 Python 2.7 and 3.6
- [x] Tested on Windows, 8.1

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -839,5 +839,5 @@ class TexTextElement(inkex.Group):
         so for full colorization of a node you need to set the fill as well as the stroke color!).
         """
         for it in self.iter():
-            if "stroke" in it.style and it.style["stroke"].lower() == "none":
+            if it.style.get("stroke", "").lower() == "none":
                 it.style["stroke-width"] = "0"
--- a/textext/base.py
+++ b/textext/base.py
@@ -395,6 +395,8 @@ class TexText(inkex.EffectExtension):
 
                     tt_node.set_meta('jacobian_sqrt', str(tt_node.get_jacobian_sqrt()))
 
+                    tt_node.set_none_strokes_to_0pt()
+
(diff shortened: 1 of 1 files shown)


textext/base.py
```diff
@@ -818,3 +820,16 @@ def import_group_color_style(self, src_svg_ele):
...
+        """
+        Iterates over all elements of the node. For each element which has the style attribute
+        "stroke" set to "none" a style attribute "stroke-width" with value "0" is added. This
+        ensures that when colorizing the node later in inkscape by setting the node and
+        stroke colors letters do not become bold (letters have "stroke" set to "none" but e.g.
+        horizontal lines in fraction bars and square roots are only affected by stroke colors
+        so for full colorization of a node you need to set the fill as well as the stroke color!).
+        """
+        for it in self.iter():
+            if "stroke" in it.style and it.style["stroke"].lower() == "none":
```
sizmailov, line 834:
```suggestion
            if it.style.get("stroke", "").lower() == "none":
```

sizmailov (approved):

textext/base.py
jcwinkler: :+1:

Issue #265, opened 2020-11-10
Double backslash (\\) when editing text from previous versions


orengt:
I just installed Inkscape 1.0.1 on an new computer with TexTex 1.2.0.
When I open existing TexTex object an additional backslash is added to all commands.
For instance in the attached file I get $\\beta$ instead of $\beta$.
[em00001-0011a Two Pending Balls.zip](https://github.com/textext/textext/files/5520425/em00001-0011a.Two.Pending.Balls.zip)

**Screenshots**
![image](https://user-images.githubusercontent.com/31801615/98737681-375ca700-23af-11eb-8075-bf74c5c5f041.png)

**System information**

- TexText version: 1.2.0
- Inkscape version: 1.0.1
- Operating system: Windows 10, Version 10.0.19041 Build 19041, 64-bit]
- Windows only: LaTeX-distribution (MiKTeX, TeX-Live): MiKTeX
- Windows only: TexText installed via batch script or installer file: batch

sizmailov:
Hi, thanks for the report! 

I recently saw similar thing on my documents: slashes were replaced by double slashes and new lines replaced by `\n` (two chars). Not sure were it comes from. At the moment of TexText object creation everything seems to be OK. My guess is that inkscape somewhere does "normalization" of attribute values. In the file you attached the value of text is already has double slashes, so it's not really helpful :(

I'm afraid we can't do much to already corrupted nodes. I'm curious where this corruption comes from :thinking: 

Note: space characters in xml attributes [are recommend to be replaced](https://www.w3.org/TR/1998/REC-xml-19980210#AVNormalize) by whitespace, so plain text attribute is not sufficient to keep arbitrary user input intact. 

@jcwinkler I think we should (optionally?) store base64-encoded text input in addition to plain `text` field to avoid this kind of problems. See #266 for details.

sizmailov:
Ok, my bad. It's actually problem of textext :disappointed: The difference comes from #242 which removes escaping difference for different python interpreters but introduces discrepancy between textext versions.

Looks like we don't store enough meta information to 100% reliably restore corrupted strings. Probably guess-and-fix could be good enough?

orengt:
Hi Sergei 

It's amazing how quickly you answered. 
I just don't understand if your question "Probably guess-and-fix could be good enough?" was for me, if it is to me, could you please explain it for me I don't think I get it.

Thanks,
Oren

sizmailov:
[...]

PR #266, opened 2020-11-11
Store node text in base64 (in addition to plain string)

sizmailov:
This PR add extra field `data-base64-text` to TexText nodes with base64-encoded `text` field. 
Encoded value is hardened against legit transformations over svg attributes (which we cannot control) and helps us keep user input intact, unlike plain text attribute. 

Plain text `text` attribute is kept for backward compatibility and used as fallback.

Related issue(s):
#265

Short checklist:
- [X] Tested with Inkscape version: [Inkscape 1.1-dev (cfef7036a0, 2020-11-10)]
- [X] Tested on Linux, Distro: [Ubuntu 20.04.1 LTS]

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -40,6 +40,7 @@ Florent Becker and Vladislav Gavryusev for contributions.
 """
 
 from __future__ import print_function
+import base64
 import hashlib
 import logging
 import logging.handlers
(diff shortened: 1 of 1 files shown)


sizmailov:
Here is and example when this PR becomes handy. 
Complex formulae usually span multiple rows, especially when it comes to matrices, etc. 

<details><summary>Before/After</summary>

Formula:

image

Before:
image

After:
image

</details>

jcwinkler:
Thanks, @sizmailov for the quick response. I will merge this and make a bugfix release during this day.

jcwinkler:
@sizmailov Could you rebase your commit onto develop, please? Thanks!

sizmailov:
Hi. I'm now thinking that I was too eager with this PR. May be we should revert to string escaping scheme instead? It has no problems with spaces/new lines and would not bloat the resulting SVG. 

Anyhow this PR won't solve #265, there is no rush with it. I'll turn it into a draft for the moment.

PR #271, opened 2020-11-18
Don't copy temporary files on install

sizmailov:
Python creates a bunch of temporary files which are copied on installation. Might be useful for devs only.

Short checklist:
- [X] Tested with Inkscape version: [fInkscape 1.1-dev (ad8effaa6e, 2020-11-17)]
- [X] Tested on Linux, Distro: [Ubuntu 20.04.1 LTS]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ import shutil
 import sys
 import stat
 import tempfile
+import fnmatch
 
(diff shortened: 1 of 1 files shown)


jcwinkler:
[x] Tested on Windows, Version 10 2004 + Inkscape 1.0.1

PR #272, opened 2020-11-22
CI Tests for Inkscape 1.0.1 + TexText 1.1/1.2

jcwinkler:
Content:

- Setup our usual tests for TexText 1.1 and 1.2 on Inkscape 1.0.1
- Added a new test (number 10) which fails on issue #265 
- For this purpose a small change in the TexText command line API has been necessary: If TexText is called with the option `--text=""` it just recompiles the specified node in the document. By this approach we can test if TexText is able to correctly read back and process text saved in a previous version (commit 6e7c66e)
- The output of the svg document on the console has been removed

On Kubuntu 20.04 all tests succeed with TexText 1.3.0-dev.1 and Inkscape 1.0.1.

Test `inkscape-1.0.1/textext-1.1.0/inkscape/10` fails when run against tag `1.2.0` (without fix 2a262c4) - as expected.

Open topics:

- [x] Fix Ubuntu 18.04 / Python 3.6 issue (see failed CI test)
- [x] Check Python2 compatibility ~~(Inkscape's standard interpreter on Ubuntu 18.04)~~ (tested TexText manually against Python2 from PyCharm, `compatibility_test.py` does not handle utf8 string correctly comming from the `config.json. I do not think it is worth spending much time on this, since, as I found out, Inkscape 1.x uses python3 be default also on Ubuntu 18.04)


sizmailov:
This seems to be quite a lot of work :muscle:  

I noticed that travis output contains (harmless?) [error messages](https://travis-ci.com/github/textext/textext/jobs/445442409#L731-L732) and unnecessary full svg printouts. Both should be avoided.

snippets/inkscape-1.0.1/textext-1.1.0/inkscape/06/config.json
```diff
@@ -0,0 +1,24 @@
...
+		"preamble-file" : null
+	},
+	"check" : {
+		"render" : {
+			"scale-is-preserved": true,
+			"render-area": "document",
+			"dpi": 300
+		},
+		"compare" : {
+			"fuzz" : "100%"
```
sizmailov, line 21: Why fuzz is 100%?

snippets/inkscape-1.0.1/textext-1.2.0/inkscape/06/config.json
```diff
@@ -0,0 +1,24 @@
...
+		"preamble-file" : null
+	},
+	"check" : {
+		"render" : {
+			"scale-is-preserved": true,
+			"render-area": "document",
+			"dpi": 300
+		},
+		"compare" : {
+			"fuzz" : "100%"
```
sizmailov, line 21: ditto

textext/base.py
```diff
@@ -305,8 +305,13 @@ def preview_callback(_text, _preamble, _preview_callback, _tex_command, _white_b
...
             else:
-                # ToDo: I think this is completely broken...
-                self.do_convert(self.options.text,
+                # In case TT has been called with --text="" the old node is
+                # just re-compiled if one exists
+                if self.options.text == "" and text is not None:
+                    new_text = text
+                else:
+                    new_text = self.options.text
+                self.do_convert(new_text,
```
sizmailov, line 314: I would prefer cryptic magic value to new option (e.g. `--recompile`), but it's not a public interface of the plugin and serves for tests only, so I guess it's not worth extra efforts.

snippets/inkscape-1.0.1/textext-1.1.0/inkscape/06/config.json
jcwinkler: This is `nonsense="100%"` of course. I fixed it.

textext/base.py
jcwinkler: I agree.

jcwinkler:
I adjusted the fuzz factor for test number 06 (the rotation test). Under Ubuntu 18.04 quite large values are required. But after a manual visual check I can say that it is ok to do so. I guess that different  imagemagick versions between ubuntu 18.04 (->6.9.7) and ubuntu 20.04 (-> 6.9.10) may lead to different pixel comparison results.

However, carefully inspecting all tests again I noticed that under ubuntu 20.04 all tests end up with exactly 0 pixel difference. That made me skeptical. I then setup the tests such that every node is replaced by `xyz` so all tests except number 10 should fail.This happens on ubuntu 18.04 but NOT on ubuntu 20.04. There, only colored nodes failed in the comparison.

Please see the attached four files. Running `compare -metric ae 1.png 2.png diff12.png` results in 0px difference which is obviously wrong. On the other hand, `compare -metric ae 3.png 4.png diff34.png` leads to a large pixel difference (1.png and 2.p
ng are from test 01, 3.png and 4.png from test 03).

Do you have any idea what I missed?

`compare -metric ae 1.png 2.png diff12.png`



**Results in empty diff:**


`compare -metric ae 3.png 4.png diff34.png`



**Results in expected diff:**

sizmailov:
I confirm, `compare` from `ImageMagick 6.9.10-23 Q16 x86_64 20190101` on ubuntu 20.04 returns zero (installed from repo).

Seems like we faced https://github.com/ImageMagick/ImageMagick/issues/1263. 
I've tested recent `ImageMagick 7.0.10-43 Q16` (build from sources) and it seems to work as expected (`compare -metric ae 1.png 2.png diff12.png` returns 3325). The commit message from latest comment in the linked issue suggests that it should be fixed starting from `6.9.10-11`, but apparently is not quite. 

Not sure if we can do much about it in CI apart from downloading patched imagemagick binaries from somewhere...

jcwinkler:
Thanks for investigating this and posting to the imagemagick devs. 

Meanwhile I successfully performed the tests with Imagemagick 7 on Windows. Doing so, I discovered some other (Windows specific) errors which I fixed.

I would merge this now and release 1.3.0 if you agree.

sizmailov:
Yeah, looks good to me!

jcwinkler:
I updated our test function according to the answer given in https://github.com/ImageMagick/ImageMagick/issues/1263.

(On my local machine I have problems with the cyrillic alphabet now, will investigate later)

jcwinkler:
@sizmailov Hmm, something is strange. Actually it does not relate to the last commit, but, please, try this on your machine:

LaTeX-Snippet:
```latex
\documentclass{article}

\usepackage[utf8]{inputenc}
\usepackage[T2A]{fontenc}
\usepackage{amsmath,amsthm,amssymb,amsfonts}
\usepackage{color}

\begin{document}
Ильюшин Ил-62 - элегантный самолет!
\end{document}
```
Save it as `test.tex`.

Then run
```bash
pdflatex test.tex -interaction=nonstopmode -halt-on-error
```
One obtains the following pdf:
grafik

Then, as TexText does, convert to png:
```bash
inkscape --pdf-poppler --pdf-page=1 --export-type=svg --export-text-to-path --export-area-drawing --export-filename test.svg test.pdf
```

Result:
grafik


```bash
> inkscape --version
Inkscape 1.0.1 (1.0.1+r74)
    Pango version: 1.44.7
```

Obviously, TexText is affected by this problem (tested with 1.1, 1.2 and 1.3-dev.1) and all Tests having this type of input fail (tests 08 and 10). First, I thought it is a misconfiguration of my machine, but the same seems to happen on Travis (see failed tests above). No problems with the tests on Ubuntu 18.04 and MS Windows.

Can you check it on your installation? Do you have any idea? Thanks!

jcwinkler:
OK, `cm-super` was missing.

This has never been discovered by me on Ubuntu 20.04 because failing test was hidden by the imagemagick issue. :angry: 

Sorry for the noise.

sizmailov:
Sorry, I wrote a comment about `cm-super` package hours ago and forgot to hit send button :facepalm:

sizmailov (approved):

PR #274, opened 2020-12-17
Update documentation reflecting Debian packaging

aerusso:
Skipping the checklist, since it isn't really relevant.  This merge simply updates the documentation reflecting the presence of the inkscape-textext package in both [Debian](https://packages.debian.org/source/testing/inkscape-textext) and [Ubuntu](https://launchpad.net/ubuntu/+source/inkscape-textext).

This doesn't update the readme, maybe that should be updated as well?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,13 +30,11 @@ Download
 
 Download of the most recent version (and older ones): https://github.com/textext/textext/releases
 
-
 Documentation
 -------------
 
-Documentation hosted at https://textext.github.io/textext
-It contains `installation <installation-toc_>`_ and `usage <usage-toc_>`_ instructions
-
+Documentation hosted at https://textext.github.io/textext.
+It contains `installation <installation-toc_>`_ and `usage <usage-toc_>`_ instructions.
 
 History
 -------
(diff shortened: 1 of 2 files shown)


sizmailov:
Hi!

Thanks for packaging efforts! That's a prominent step forward for TexText)

The change looks good to me. Although I'm not sure _when_ we should publish that. Current Ubuntu 20.04 [doesn't have the package](https://packages.ubuntu.com/search?suite=all&arch=any&searchon=names&keywords=textext) (yet?) and it would confuse users at the moment. I think same applies to Debian (didn't check). Either we should list distributions that are capable of installation via package manager or postpone publication of the instructions. 

Minor question. Do you think it's worth to mention system-wide/per-user installation difference between two (and as consequence no need for `sudo` in latter case)?

aerusso:
I agree about the timing concern.  Debian users will certainly understand if they are using Bullseye "or later," but downstream distributions may not.  Also, I do not have an Ubuntu machine to test on, so I don't quite know how Ubuntu's decision to replace the Inkscape `.deb` with the snap will affect things---I suspect it will break it, per the existing documentation.  I don't quite know, though, if the Inkscape PPA will work---again, I suspect it will, but I don't have a machine to test it on.

As for the "system" vs "per-user" distinction: maybe add language like "TexText can be installed **system-wide** directly from the official repositories..." and "If not, continue to follow the manual, **per-user** installation instructions."

jcwinkler:
Sorry, merged it in wrong branch. Moved it manually to develop now.

Issue #280, opened 2021-02-02
the setup_win.bat _requires_ double quotes on pathname, this isn't clear


PinkMushroom:
First of all, i had a freshly installed inkscape using the msi and the batch file couldn't find it, but that's a different issue. 
However, not a problem, I'll just specify the installation path, thusly

`setup_win.bat --inkscape-executable C:\Program Files\Inkscape\bin\`

and then got this message

> Trying to detect Python interpreter in Inkscape installation directory...
> No Python interpreter found within your Inkscape installation.
> Expected as "C:\Users\xxxxxx\Downloads\textext-1.3.0\Files\Inkscape\\python.exe"
> You have to install Inkscape with the Python option enabled
> in the Inkscape installer.

well that's interesting because i absolutely did enable python.

Then i noticed this ;

> Specifiy an explicit directory via the --inkscape-executable option
> to look for if you installed Inkscape from a zip package. E.g.:
> setup_win --inkscape-executable "C:\Path\to\Inkscape installation\"

maybe i need double quotes !

`setup_win.bat --inkscape-executable "C:\Program Files\Inkscape\bin\"`

```
"C:\Program Files\Inkscape\bin\" found
Trying to detect Python interpreter in Inkscape installation directory...
"C:\Program Files\Inkscape\bin\\python.exe" found

Trying to run "C:\Program Files\Inkscape\bin\\python.exe" "C:\Users\xxxxxx\Downloads\textext-1.3.0\setup.py" --inkscape-executable "C:\Program Files\Inkscape\bin\"...

[TexText][ERROR   ]: Bad `inkscape` executable provided: `C:\Program Files\Inkscape\bin"`. Abort installation.
```

huh that's weird.   ohhhhh. i bet it's not the directory i need but the actual path to the executable, like so

`setup_win.bat --inkscape-executable "C:\Program Files\Inkscape\bin\inkscape.exe"`

success... but at every step the error message had nothing to do with what was actually wrong. in particular the batch file is specifically requesting a directory but it's the full path to the executable that is required.

jcwinkler:
Thanks for the report!

> using the msi and the batch file couldn't find it

I will check this since the batch script is looking for the registry keys which should be set by the msi installer.

I agree that the help text

> Specifiy an explicit directory via the --inkscape-executable option
> to look for if you installed Inkscape from a zip package. E.g.:
> setup_win --inkscape-executable "C:\Path\to\Inkscape installation"

[...]

PR #284, opened 2021-02-23
Windows installer: Better handling of MSI Inkscape installations

jcwinkler:
When Inkscape is installed on Windows using the MSI installer the
registry keys are not set as it is done when using the NSIS installer.
In fact, no valid registry keys are set from which anything about the
install location can be concluded. Hence a simple trial and error method
has been added to the batch file script to resolve this in the most
likely cases.

The problem also occured as a side effect in #280 

Short checklist:
- [x] Tested with Inkscape version 1.0.2
- [x] Tested on Windows 10 2020H2

### DIFF
--- a/setup_win.bat
+++ b/setup_win.bat
@@ -13,8 +13,8 @@ rem Allow access to modified variables in if and for constructs by
 rem !VAR! (%VAR% would not result in modified value as one would expect...)
 setlocal EnableDelayedExpansion
 
+set INKSCAPE_EXENAME=inkscape.exe
 set "INKSCAPE_DIR="
-set "INKSCAPE_EXE="
 set "PYTHON_ARGS="
 set "PATHON_EXE="
 set "PYTHON_COMMAND="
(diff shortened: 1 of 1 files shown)


jcwinkler:
I also opened an issue here: https://gitlab.com/inkscape/inkscape/-/issues/2237

setup_win.bat
```diff
@@ -138,35 +139,51 @@ for %%R in (HKLM HKCU) do (
...
-			rem for %%S in ("!INKSCAPE_EXE!") do set INKSCAPE_DIR=%%~dpS
 		)
 	)
 )
 
+rem If we did non succeed in the registry lets have a look 
+rem at the most common install locations
+echo Trying the usual Windows install locations...
+for %%D in (C, D, E, F, G, H) do (
+	for %%F in ("Program Files", "Program Files (x86)") do (
```
sizmailov, line 168: I suggest to rename one-letter variables. The bigger code lifetime and variable scope the more chances it should be named better.

setup_win.bat
```diff
@@ -138,35 +139,51 @@ for %%R in (HKLM HKCU) do (
...
+		set INKSCAPE_DIR=%%D:\%%~F\Inkscape\bin
+		echo    Trying !INKSCAPE_DIR!...
+		set INKSCAPE_EXE=!INKSCAPE_DIR!\inkscape.exe
```
sizmailov, line 171: `INKSCAPE_DIR` and `INKSCAPE_EXE` seems to be used as output variables and temporaries at the same time.
It would be better to set temporaries (aka result candidates) values to something else and assign to output variables only once. I suspect there is a potential bug when those variables are set (as temporaries), test failed and they left in "FOUND" state. Could be wrong, though.

sizmailov (commented):
I didn't run the code, just read through. 

I think it's fine to add more guess&try entries to inkscape location search procedure in the time of poor guaranties provided by the inkscape installers. 

I added a couple of comments. They are a bit generic, but you still might want to address them.

setup_win.bat
jcwinkler: Unfortunately, in Windows batch only single letter variables are allowed for the iteration variable in for-loops. One of the strange things in this language...

setup_win.bat
jcwinkler: I agree. I will remove the INKSCAPE_EXE variable.

setup_win.bat
sizmailov: Didn't know that... Indeed a surprising feature :confused:

Issue #276, opened 2021-01-06
TexText does not start after MacOS Big Sur installation


Prototypemax:
**Description of the problem/ steps to reproduce**
After MacOS Big Sur installation I was not able to launch TexText from Inkscape anymore. A error box from Inkscape appears and TexText won't open. Before Big Sur Installation everything went really nice.

**Message(s) in error dialog (if this comes up)**
The error says:
"macOS 11 or later required !"
My system runs 11.1 (20C69).

**Logfile**
Please also *attach* the TexText log file (since TexText 0.9.x). You find it in `~/.config/inkscape/extensions/textext` on Linux/ MacOS or in `C:\Users\[Your user name]\AppData\Roaming\inkscape\extensions\textext` on Windows
Unfortunately I can't find the folder.

**Screenshots**
If applicable and helpful, add screenshots to help explain your problem.
![Bildschirmfoto 2021-01-06 um 19 48 47](https://user-images.githubusercontent.com/35840778/103808391-33b86a00-5058-11eb-88a6-2ba9d19cd67f.png)
Text translated: "The called script has passed additional data to Inkscape. No error message was returned by the script, but the result may be unusable."

**System information**

- TexText version: as I cannot find the folder, probably 1.2 oder 1.3. I downloaded it this autumn.
- Inkscape version: 1.0.1
- Operating system: MacOS Big Sur 11.1 (20C69)

jcwinkler:
Unfortunately, without the logfile it is very difficult to determine the source of the problem. 

`.config/inkscape/extensions/textext/textext.log` should be in your homedir. Maybe you have to turn on display of hidden files in your file explorer to see it?

I also tried to grep the error message `macOS 11 or later required` and parts of it in TexText and Inkscape source and I did not find anything.

mattfack:
You can find the log under `/Users/yourusername/Library/Application Support/org.inkscape.Inkscape/config/inkscape/extensions/textext` where you can see what is the error. In my case it does not find xelatex, lulatex or pdflatex in the path, which is strange since they are there...

mattfack:
In my case even with the line

`python3 setup.py --pdflatex-executable=$(which pdflatex) --lualatex-executable=$(which lualatex) --xelatex-executable=$(which xelatex) --skip-requirements-check`

it still fails due to GTK3 missing.

Downgrading to inkscape 1.0.0 did it, however I still get some errors when I write text, namely:

    
[...]

PR #285, opened 2021-02-25
Improvement of documentation (setup script, Debian packages)

jcwinkler:
This PR addresses three improvements in the documentation:

- The messages of the Windows setup script `setup_win.bat` are more precise now and do clearly state and illustrate what is meant. This addresses issue #280

- The command line options of the setup script are explained in detail and are associated with the corresponding setup scenarios/ problems. This addresses issue #276 (see https://github.com/textext/textext/issues/276#issuecomment-761868727). 

  Here is a rendered version: [Setup-Script options](https://jcwinkler.github.io/textext/usage/troubleshooting.html) referenced in [Linux](https://jcwinkler.github.io/textext/install/linux.html#linux-install-textext), [Windows](https://jcwinkler.github.io/textext/install/windows.html#windows-install-textext), and [Mac](https://jcwinkler.github.io/textext/install/macos.html#macos-install-textext) installation instructions.

- The possibility to install TexText packages via apt-get in Debian Bullseye has been moved into a separate section (I prefer this way as long as this is not available to a broader set of distributions. Later we can instruct sth. like "First try your package manager, if this fails go to the manual installation")

  Here is a [rendered version](https://jcwinkler.github.io/textext/install/linux.html#linux-install-textext)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,8 +21,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'TexText'
-copyright = '2020, Alexander Blinne, Brian Clarke, Florent Becker, Jan Winkler, Pit Garbe, Pauli Virtanen, Robert Szalai, Rafal Kolanski, Sergei Izmailov, Toru Araki, @veltsov, Vladislav Gavryusev'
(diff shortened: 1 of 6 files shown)


sizmailov:
I would use quotes everywhere instead of escape of individual characters. 

LGTM!

PR #297, opened 2021-05-30
Convert strokes to path

### LINKED ISSUE #291: Convert line elements to objects for easy colour change
Observation: 

- When creating a formula such as $\frac{\overline{x}}{2}$, then changing colour of whole formula requires changing colour of both area and line. Alternatively, changing colour or area and then re-opening Textext interface applies area colour to all elements (one-off effect, next area colour change will not change overlines and fraction bars colour)
image

-  Also, scaling of formula object will not thicken fraction bar and overline if scaling of lines is not activate in inkscape, which can be annoying if formula is part of a group one wants to scale without scaling of line thickness.

Analysis: The rendering tools, when converting formulae, converts at least fraction bars and overlines to lines, whereas the other symbols in formulae are borderline-free filled paths. 

Feature request: apply a transform similar to Inkscape's Path > Stroke to Path command to all shapes generated by the renderer.

jcwinkler:
Strokes are converted to path after svg generation from pdf. It resolves #291. The following figures illustrate the problem solved by this fix:

grafik

grafik

The *disadvantage* is that now two command line calls of the inkscape executable are required. Please see the discussion [here](https://inkscape.org/forums/beyond/cli-pdf-import-combined-with-actions/).

If this is too time consuming we could add a checkbox to the gui enabling this feature. However, this needs additional explanations in the doc and may be difficult to understand for the unexeperienced user. What do you think @sizmailov ?

(I needed to adjust some tests, too.)

Short checklist:
- [x] Tested with Inkscape version 1.0.2
- [x] Tested on Linux, Distro: Kubuntu 20.04
- [x] Tested on Windows, Version: 10 20H2


sizmailov:
This looks very nice! 

On my machine inkscape invocation (pdf->svg) takes about 0.6 seconds. While 0.6 is a lot, it's still better than extra clicks required to fix the rendered svg, so let's keep it as is.

My only concern is compatibility of the conversion command with different inkscape versions. If I remember correctly they were reworked recently, so it might fail for someone. To circumvent this problem the extra inskcape call should be allowed to fail. In such event we should render original svg. 

Other than that looks good to me!

Note: Conversion command didn't work on dev inkscape version from Nov 2020.

jcwinkler:
@sizmailov I updated the patch such that a failed stroke to path command is silently ignored. Note that I had to modify the `exec_command` method because Inkscape CLI does not return a non zero error code if illegal verbs are used.

I will do some further tests with Inkscape 1.0.0, 1.0.1 and most recent development versions of Inkscape.

(CI tests completely fail at the moment on ubuntu 20.04, not on 18.04, will investigate the reason later. Locally, all tests pass on my kubuntu system...)

jcwinkler:
Ahhh 🚀 : In Inkscape 1.1 a "Confirm Overwrite" Dialog opens when the `FileClose` verb is executed. Fixed this.

I will deactivate the tests on Travis. I do not now why they fail. They run successfully on my local system with Inkscape 1.0.2 as well as with 1.1. Since I will move the CI to Github Actions I do not want to waste my rare time with fixing this on Travis.

jcwinkler:
I updated the documentation:

- Usage of GUI: https://jcwinkler.github.io/textext/usage/gui.html#colorization-of-the-output
- Manual Toolchain: https://jcwinkler.github.io/textext/usage/troubleshooting.html#manual-use-of-the-toolchain

After having tested the patch on Inkscape 1.0.2 and 1.1.0 on Windows 10 and Kubuntu 20.04 I open this PR of final review @sizmailov

sizmailov:
> Note that I had to modify the exec_command method because Inkscape CLI does not return a non zero error code if illegal verbs are used.

I think this is unnecessary. Here we don't care whether it succeeded or not, we use same filename after all.

Tested on kubuntu 20.04 with Inkscape 1.2-dev (dc52390971, 2021-05-29).

sizmailov (approved):

PR #304, opened 2021-07-21
Opt-in possibility for stroke to path conversion

jcwinkler:
In TexText 1.4.0 I introduced the default conversion of strokes to path to ensure proper and easy coloration in Inkscape, see #297.

I am pretty annoyed by this feature in my daily work because it consumes too much time. Additionally, under Windows a short flickering appears during the extra invocation of an Inkscape binary (this is required by the operation).

Hence, I added a checkbox which allows the user to opt in this feature. I intentionally use an "opt-in" approach since I think colorization of formulae is not that often used. 

image

Things to be done:
- [x] Enlarge TeX code edit box again
- ~~[ ] Add option to GUI for default behavior~~
- [x] TK support (?)
- [x] Update documentation

Short checklist:
- [x] Tested with Inkscape version: Inkscape 1.1 (1:1.1+202105261517+ce6663b3b7)
- [x] Tested on Linux, Distro: Kubuntu 20.04


sizmailov:
I don't like change of UI layout. Checkbox seems to be of rare use while it takes plenty of space right in center of the window. I would move it to `Settings`.

jcwinkler:
I updated the geometry of the Window such that the view is more balanced now and the code is in the center.

I would like to keep the checkbox in the main window for the following reasons: 

1. The settings on the main Windows refer to the appearance and behavior of the produced output. The settings in the settings menu refer to the appearance and behavior of the GUI only. Hence, I find it logical to keep the stroke to path option grouped together with the other node related settings.

   (If this distribution of settings is reasonable or not is another question and should be discussed if anyone has the time to implement the resulting changes...)

2. It is in a prominent place so a user might be aware of this possibility without going through the docs and menus.

sizmailov:
As a compromise checkbox can be framed in a box with shorter title, e.g. "No strokes" without extra text nearby and moved next to "Alignment" frame (keeping checkbox hover text).

Usability is directly coupled to usage frequency of button/menu item, therefore I don't see a problem with moving it in sub-menu. Anyhow it's a matter of personal preference, so I won't step into any further discussion.

jcwinkler:
New proposal:

image

jcwinkler:
Updated doc: https://jcwinkler.github.io/textext/usage/gui.html#colorization-of-the-output

Issue #314, opened 2021-09-17
Extension tries to write to installation directory instead of defaults.inkscape_extensions_path again (regression)


Thermi:
**Before submitting a bug report make sure you have checked the
  [FAQ](https://textext.github.io/textext/usage/faq.html) and
  [Troubleshooting](https://textext.github.io/textext/usage/troubleshooting.html)
  sections in the documentation!**

**Description of the problem/ steps to reproduce**
A clear and concise description of what the bug is and what you expected to happen.

I received a user report that #111 is back. So textext tries to write to the installation directory again, instead of `defaults.inkscape_extensions_path` that the MR (#113) that fixed #111 used.

User report: https://aur.archlinux.org/packages/textext/

jcwinkler:
Thanks for the report. I confirm this. This bug (most likely to be relevant only if a global installation is used) has been introduced by commit 214d70eec442dfc7d568ae3caef0d242da727ee9 and PR #308.

Background is the preparation of TexText to be used in the extension manager. When TexText is installed via the extension manager `defaults.inkscape_extensions_path` is not the path that is used when it is installed manually via setup. Unfortunately I simply forgot that this breaks a global installation. 😞 

I re-think this approach to find a better solution. Maybe it is better write the config into `~/.config/textext` or `C:\Users\[USERNAME]\AppData\config\textext` and store this in a new variable `defaults.textext_config_path` ?

jcwinkler:
@Thermi Maybe you could stop packaging version 1.5.0 and use 1.4.0 instead meanhile? In 1.4.0 config is written to `defaults.inkscape_extensions_path`.

PR #315, opened 2021-09-19
base: Rework logging location

Thermi:
So the order is script location, inkscape_extensions_path, cwd
Fixes issues with non-writable log location

Related issue(s): #314

Tries to write first into script locatin, then inkscape_extensions_path, then the current working directory.
It sets up the user channel first so in any case the logs are accessible if you decide to later show the logs if
using/setting up the file logger fails

Not tested yet, should work fine though.

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -33,10 +33,6 @@ EXIT_CODE_OK = 0
 EXIT_CODE_EXPECTED_ERROR = 1
 EXIT_CODE_UNEXPECTED_ERROR = 60
 
-LOG_LOCATION = os.path.join(os.path.dirname(__file__))
-if not os.path.isdir(LOG_LOCATION):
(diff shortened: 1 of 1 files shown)


jcwinkler:
Thanks for the work, @Thermi Meanwhile I am also working on an implementation relating this issue (see [this branch](https://github.com/jcwinkler/textext/tree/feature/global_installation)). It will also enable a global installation from the command line.

I prefer to have the logfile/ settings in one location, independent of the way how TexText has been installed (via setup script locally or via package globally or via the extension manager).

Thermi:
Sure. It's entirely your choice. I'd consider this though. Eventually with a log channel over stderr, to which we can then log error messages when previous log location candidates are not accessible, and to log the final file log location to.

PR #330, opened 2021-11-30
fix no pango package error

xuestrange:
I think `pango` is not an independent package anymore, I have to import `Pango` from `gi.reposity` which can be installed via `pip install PyGObject`. Then `set_monospace_font`  function works well and you can change the font and fontsize of the texts you input, it's more eye-comfortable.

I‘m sorry that I don't have a MacOS device in hand.
Jetbrains Mono, 18

Short checklist:
- [x] Tested with Inkscape version: Inkscape 1.1.1 (3bf5ae0d25, 2021-09-20, custom)
- [x] Tested on Linux, Distro:  x86_64 Linux 5.4.159-1-MANJARO
- [x] Tested on Windows, Version: Windows 10 21H1


jcwinkler:
Great, looks good. Let me do some additional tests, then I will merge it. Thanks!

jcwinkler:
Thanks @xuestrange for this contribution. Additionally, I will add an entry in the settings menu allowing to choose the font out of the set (10, 11, 12, 14, 16, 18).

PR #339, opened 2022-01-03
Feature/user units inkscape 1.2

### LINKED ISSUE #338: Inkscape 1.2-dev: New nodes are not placed in the center of the view
**Description of the problem/ steps to reproduce**
When running TexText in the **development** version of Inkscape **1.2** together with the Inkscape extension system from the extension **master** branch new nodes are not placed in the center of the view. Instead, they are placed far away from the current view (usually bottom right, you need to zoom out to find it).

**Screenshot**
The node has been created while the A4 pager was filling the view completely (Shortcut `5`). It should have been placed in the center of the page. In the screenshot it became visible after zooming out to 28%
grafik

**System information**

- TexText version: At least >= 1.6.0
- Inkscape version: Inkscape 1.2-dev (0677221, 2021-12-22)
- Operating system: all

jcwinkler:
Related issue(s): Resolves #338 

Pull request addresses the fact that in Inkscape 1.2. the extension API method [`unittouu`](https://inkscape.gitlab.io/extensions/documentation/source/inkex.elements._base.html#inkex.elements._base.BaseElement.unittouu) always converts a value into px and not into the units specified by the user. Hence, the view_center is always returned in px now and not in the units selected by the user in its document. Depending on the API version the view_center coordinates are converted into the correct unit now.
```python
# Since Inkscape 1.2 (= extension API version 1.2.0) view_center is in px,
# not in doc units! Hence, we need to convert the value to the document unit,
# so the transform is correct later.
if hasattr(inkex, "__version__"):
   if check_minimal_required_version(inkex.__version__, "1.2.0"):
   view_center.x = self.svg.uutounit(view_center.x, self.svg.unit)
   view_center.y = self.svg.uutounit(view_center.y, self.svg.unit)
```
([Code reference...](https://github.com/jcwinkler/textext/blob/311bd762292178227a0b7de0ef45d78ebdbc77f7/textext/base.py#L395-L401))

I also reworked the scaling of the poppler output. It is not scaled in mm now but in the units used by the user in the document:
```python
# Ensure that snippet is correctly scaled according to the units of the document
# We scale it here such that its size is correct in the document units
# (Usually pt returned from poppler to mm in the main document)
self.transform.add_scale(root.uutounit("1{}".format(root.unit), document_unit))
```

([Code reference...](https://github.com/jcwinkler/textext/blob/311bd762292178227a0b7de0ef45d78ebdbc77f7/textext/base.py#L667-L670)...)

I tested it with documents in differnt units (inch, mm, px) and it seems to work now.

Further changes:

- Some test snippets
- Helper function `check_minimal_required_version` in `utility.py`

@sizmailov Would you agree with this approach? 

Short checklist:
- [x] Tested with Inkscape version: 1.1 and 1.2-dev
- [x] Tested on Linux Kubuntu 20.04
- [x] Tested on Windows 8.1 and 10 21H2

### DIFF
--- a/textext/utility.py
+++ b/textext/utility.py
@@ -20,6 +20,7 @@ import shutil
 import stat
 import subprocess
 import tempfile
+import re
 
 from .errors import *
 import sys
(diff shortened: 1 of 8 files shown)


textext/utility.py
```diff
@@ -259,6 +260,38 @@ def exec_command(cmd, ok_return_value=0):
     return out + err
 
 
+def check_minimal_required_version(actual_version_str, required_version_str):
```
sizmailov, line 263: A better name for this function is `version_greater_than`. The "actual" and "minimal"/"required" nature of arguments passed by  ***a** caller* of the function is unrelated to the function definition.

sizmailov (approved):
Looks good to me! (Didn't test, though)

PR #340, opened 2022-01-14
Allow optional \documentclass or \documentstyle in preamble

PatrickKa:
Related issue(s): none

**Reasoning**: My reason for this PR is that I want to create figures for a document where the normal font size is set to 11pt (instead of the default 10pt) and I do not want to manually specify that font size in every textext cell. The proposed change allows to have a `documentclass`-like command in the preamble which makes it easy to "globally" specify things like the normal font size. The change does not alter the behavior for preambles without such a command. It is fully backwards compatible.

**Implementation**: I moved `\documentclass{article}` from the `DOCUMENT_TEMPLATE` into its own constant called `DEFAULT_DOCUMENT_CLASS`. This constant is only prepended to `preamble` if it does not contain a `\documentclass` or `\documentstyle` command. To check this I added the function `_contains_document_class(preamble)` which parses the preamble for those commands and also makes sure that they are not commented out.

Short checklist:
- [x] Tested with Inkscape version: 1.0.2-2
- [ ] Tested on Linux, Distro:
- [x] Tested on Windows, Version: Windows 10 Home (19042.1466)
- [ ] Tested on MacOS, Version:

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -505,8 +505,8 @@ class TexToPdfConverter:
     """
     Base class for Latex -> SVG converters
     """
+    DEFAULT_DOCUMENT_CLASS=r"\documentclass{article}"
     DOCUMENT_TEMPLATE = r"""
-    \documentclass{article}
     %s
     \pagestyle{empty}
     \begin{document}
@@ -542,6 +542,10 @@ class TexToPdfConverter:
             if os.path.isfile(preamble_file):
                 with open(preamble_file, 'r') as f:
                     preamble += f.read()
+
+            # Add default document class to preamble if necessary
+            if not _contains_document_class(preamble):
+                preamble = self.DEFAULT_DOCUMENT_CLASS + preamble
 
             # Options pass to LaTeX-related commands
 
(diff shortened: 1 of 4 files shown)


jcwinkler:
Thank you for sharing this @PatrickKa. I think it is good to add such kind of flexibility. We should add this to the documentation, as well. Or maybe - better - ship the default preamble with the documentclass definition inside of it in future version (together with your fix for backward compatibility).

textext/base.py
```diff
@@ -636,6 +640,22 @@ def parse_pdf_log(self):
                 return "TeX compilation failed. See stdout output for more details"
 
 
+def _contains_document_class(preamble):
+    """Return True if `preamble` contains a documentclass-like command.
+    
+    Also, checks and consideres if the command is commented out or not.
```
jcwinkler, line 646: Please fix typo -> `considers`

PatrickKa:
I changed the default preamble file as you suggested and updated the documentation accordingly. I did not update the documentation under `textext_0x`. I hope that this is fine.

jcwinkler:
Thanks, this is really great work. I release this as 1.8 next weekend. 

> I did not update the documentation under textext_0x. I hope that this is fine.

This is fine since TexText 0.x is deprecated and for Inkscape 0.92.x only.

PR #391, opened 2023-07-25
Fix: GIO::Error

### LINKED ISSUE #390: Inkscape 1.3: TexText reports Gio::Error
**Steps to reproduce the behavior (if not obvious from description)**:
1. Start Inkscape 1.3 and create new document
2. Run TexText extension 
3. Type `$$1$$` in TexText dialog window
4. Press `Preview` button

**Actual Result**
_Error occurred while generating preview_ window pop-ups

**Expected Result**
Rendered `1` appear in the bottom of TexTex dialog window. 

**Message(s) in error dialog**

<details>
<summary>dialog output</summary>

```
Command C:\Program Files\Inkscape\bin\inkscape.exe --pdf-poppler --pdf-page=1 --export-type=png --export-area-drawing --export-dpi=300 --export-filename tmp.png tmp.pdf failed (code 3)
```
</details>

<details>
<summary>output of stderr:</summary>

```
terminate called after throwing an instance of 'Gio::Error'
```
</details>

**System information**

- TexText version: 1.8.2
- Inkscape version: Inkscape 1.3 (0e150ed6c4, 2023-07-21)
- Operating system: Windows 10, 22H2, 64-bit]
- Windows only: LaTeX-distribution MiKTeX
- Windows only: TexText installed via batch script

**Additional information**

I did not fresh install of Incscape and TexText. When the new version 1.3 came out, I uninstalled the old version 1.2.1 (leaving the profile in %APPDATA%) and installed the new one. The extension was also installed earlier in version 1.2.1. Thus, now version 1.3 works with the old profile from 1.2.1.

**Workaround**
You may have installend Inkscape 1.2.2 with Textext, and unzipped Inkscape 1.3 (from official 7z archive, downloaded from inkscape.org) somewhere another place . 

Since the unzipped Inkscape uses the same settings directory in %APPDATA% as the installed one, the TexText extension will be available to you in version 1.3.

Since the extension was installed for Incscape version 1.2, internally for its work it will run the inkscape.exe v1.2 for its needs.

Thus, you can use the extension while working in version 1.3. Checked.

jcwinkler:
When calling Inkscape 1.3 from the self written routine `exec_command`, an `GIO:Error` is thrown. We need these calls for version determination and pdf to svg/png conversion.

While the reason for this behaviour is currently unknown the error disappears when the inkscape function provided by the `inkex.command` module is used instead. This is implemented now.

During setup we cannot use this module. Here we still use the self written `exec_command` which works well when invoked from the setup-routine.

Resolves #390


jcwinkler:
Ok, ` https://gitlab.com/inkscape/extensions/raw/master/inkscape.extension.rng` needs to be fixed as well in the CI test.

Issue #403, opened 2023-10-19
Typst support


Atreyu-94:
Would it be possible for you to support Typst? It has priority support for svg and pdf.

https://typst.app/docs/

jcwinkler:
Yes, this should be easy to implement. I will try this. Nice idea!

(Later we can decide if this should go into a separate extension because the name TexText does not lead intuitively to Typst)

Atreyu-94:
I think it's a great idea to group them all within the same extension. Maybe making a clarification in the readme.
I've been using this extension for a long time and compatibility with Typst would be great :D

jcwinkler:
ToDo for TexText 1.10:

- [ ] If typst is selected: disable preamble file
- [ ] TeX command --> Compiler command (pdflatex, typst, ...)
- [ ] Update documenation

ToDo TexText 2.0
- [ ] typst is able to produce svg code directly. Hence we do not need pdf -> svg conversion for typst

cknoll:
I can confirm that typst support works on my system (debian 11). Because I do not have GTK support installed it automatically opens in TkInter-mode.


![Screenshot_20231026_121147](https://github.com/textext/textext/assets/1315698/c5915b0c-b35f-4cf0-98d9-17784f4d688d)

jcwinkler:
This is implemented now in release 1.10.0.

OpenFoam-User:
@jcwinkler Hi, thank you for adding typst support.

I have downloaded the exe version of textext and  installed it on Windows 11. I am using  `Inkscape 1.4-dev (4ca261c, 2023-11-03, custom)`.

When I open the textext editor, I do not see typst listed in list of compilers despite being added to System Path.

![image](https://github.com/textext/textext/assets/67183566/15b9ec1b-5217-4e0a-8802-05200e24b7e3)


**P.S.** When I run the following command on cmd or powershell: 

```
typst --version
typst 0.8.0 (360cc9b9)
```

YDX-2147483647:
> When I run the following command on cmd or powershell

@OpenFoam-User Are you on Windows? `setup_win.bat` may have explained the reason.

```log
[TexText][SUCCESS ]: /-and-+ [Succ] Detect *latex
[TexText][SUCCESS ]: |     /--or-* [Succ] `pdflatex.exe` is found at `…`
…
[TexText][INFO    ]: |     /--or-* [Fail] `typst` is NOT found in PATH
[TexText][INFO    ]: |            Please follow installation instructions at
[...]

Issue #378, opened 2022-11-24
Release 2.0 follow-up


jcwinkler:
**Release:** December 2022

Current (more or less stable) state can be found in [TEXTEXT_2_0_Dev](https://github.com/textext/textext/tree/TEXTEXT_2_0_Dev) branch, more recent developments in my fork [TEXTEXT_2_0](https://github.com/jcwinkler/textext/tree/TEXTEXT_2_0) (without any guarantee). Do not use that stuff in a productive environment.

The following list reflects the state of my local development which from time to time flows into the above mentioned branches.

### Main goals:
- Support installation via Inkscape’s extension manager
- Limit compatibility to Inkscape 1.2.x (inkex API >= 1.2.0)
- Remove Python 2.x compatibility
- Ensure Python >= 3.6 compatibility 
- General code clean up
- consistent code style

### General tasks

**Code**
- [x] Simplify/ clean-up requirements.py (we need not the complexity of the old days with pstoedit/pdf2svg etc)
- [x] Clean-up workflow in base.py (use meta data structures instead of tons of variables, etc.)
- [x] Simplify/ clean-up setup.py
- [x] Clean-up logging
- [x] Re-think what is in which module: environment – dependencies – logging
- [ ] Refactor gui code
  - [ ] more meaningful names (#161)
  - [ ] alignment icons for dark theme (#237)
  - [ ] simplify save/preview callback mechanism
- [ ] winapp_paths.py not needed for TexText (but for setup.py)

**Functionality**
- [ ] User must be able to manually define location of latex interpreters (because extension manager cannot do this)
	- [x] Create a configure dialog
	- [ ] Include this dialog into gui
- [ ] Easy way to edit preamble
- [ ] Meaningful error messages of unexpected errors (traceback)
- [ ] Add possibility to import snippet from file
- [ ] Setup via MS Store

**Check**
- [ ] Handling of preamble
- [x] No Python 2.x code anymore

**Fix known bugs**
- [ ] Strokes are not colored after re-editing node
- [ ] Images in pdf code not included (#358)

**Doc**
- [ ] User: Add known issue section
- [ ] Code: consequent rst sphinx style
- [ ] Add #153

**Pass PyLint requirements**
- [x] setup.py
- [ ] base.py
- [x] environment.py
- [x] errors.py
- [ ] gui.py
- [x] log_util.py
- [x] settings.py
- [x] utility.py

**Tests/CI**
- [ ] Check
- [ ] Remove Inkscape 1.0.1 tests
[...]

PR #406, opened 2023-11-18
Fix `typst.exe` ignored on Windows 

YDX-2147483647:
Related issue(s): https://github.com/textext/textext/issues/403#issuecomment-1817494909

On Windows, executables end with `.exe`. The setup script thought `typst is NOT found in PATH` by default, because it does not know `typst.exe`. This PR tells it to check `typst.exe`.

Short checklist:
- [ ] Tested with Inkscape version: 1.3 (0e150ed6c4, 2023-07-21)
- [ ] Tested on Linux, Distro: —
- [ ] Tested on Windows, Version: 10
- [ ] Tested on MacOS, Version:  —
- [x] Download the ZIP release of https://github.com/textext/textext/releases/tag/1.10.0 for Windows, apply this PR, and it works.

(Sorry this is a trivial PR so I didn't test rigorously)

### DIFF
--- a/textext/requirements_check.py
+++ b/textext/requirements_check.py
@@ -132,7 +132,7 @@ class WindowsDefaults(Defaults):
                         "pdflatex": ["pdflatex.exe"],
                         "lualatex": ["lualatex.exe"],
                         "xelatex": ["xelatex.exe"],
-                        "typst": ["typst"]
+                        "typst": ["typst.exe"]
                         }
 
     def __init__(self):


jcwinkler:
Good catch! Thank you.

YDX-2147483647:
Besides, [`shutil.which()`](https://docs.python.org/3/library/shutil.html#shutil.which) provides a simple and robust way to locate the executable on `PATH`.

```python
>>> from shutil import which
>>> which('pdflatex')
'…\\texlive\\2023\\bin\\windows\\pdflatex.EXE'
>>> which('typst')
'C:\\Users\\…\\scoop\\shims\\typst.EXE'
```

You can use it if you plan to refactor [`_find_executable_in_path()` etc.](https://github.com/textext/textext/blob/0d6627aeeff77d7f1758e571c44b58ad02cc6d3f/textext/requirements_check.py#L675-L690) in [v2.0](#378).

jcwinkler:
Yes, all this path detecting code will be removed in TexText 2.0 and replaced by clean solutions.

PR #442, opened 2024-12-12
Add typst preamble support

TheZoq2:
Thanks for this plugin, it makes making nice looking figures so much easier :heart: 

I was very happy to see typst support was added, and for my use case, I'd really like preamble support. I'm not sure if there is a reason for it being disabled before, but this system seems to just work for me at least

Followup to #403 I suppose

Short checklist:
- [x] Tested with Inkscape version: [1.4.1]
- [x] Tested on Linux, Distro: [Arch Linux]
- [ ] Tested on Windows, Version: [N/A]
- [ ] Tested on MacOS, Version:  [N/A]

I don't have easy access to windows or macos systems to test with

### DIFF
--- a/textext/asktext.py
+++ b/textext/asktext.py
@@ -438,9 +438,8 @@ class AskTextTK(AskText):
         self._gui_config["word_wrap"] = self._word_wrap_tkval.get()
 
     def on_texcmd_change(self):
-        using_tex = self._tex_command_tk_str.get() != "typst"
-        self._preamble["state"] = Tk.NORMAL if using_tex else Tk.DISABLED
-        self._askfilename_button["state"] = Tk.NORMAL if using_tex else Tk.DISABLED
+        self._preamble["state"] = Tk.NORMAL
+        self._askfilename_button["state"] = Tk.NORMAL
 
     def reset_scale_factor(self, _=None):
         self._scale.delete(0, "end")
@@ -819,10 +818,7 @@ class AskTextGTKSource(AskText):
         self.window_deleted_cb(widget, None, None)
 
     def cb_compiler_changed(self, combo_box):
-        using_tex = self.TEX_COMMANDS[self._texcmd_cbox.get_active()] != "typst"
-        self._preview_button.set_sensitive(using_tex)
-        self._preamble_widget.set_sensitive(using_tex)
-        self._preamble_delete_btn.set_sensitive(using_tex)
+        pass
 
     def move_cursor_cb(self, text_buffer, cursoriter, mark, view):
         self.update_position_label(text_buffer, self, view)
--- a/textext/base.py
+++ b/textext/base.py
@@ -599,27 +599,19 @@ class TexToPdfConverter:
         """
 
         with logger.debug("Converting .typ to .{0}".format(file_type)):
-            # # Read preamble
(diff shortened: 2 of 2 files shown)


textext/asktext.py
```diff
@@ -438,9 +438,8 @@ def cb_word_wrap(self, widget=None, data=None):
         self._gui_config["word_wrap"] = self._word_wrap_tkval.get()
 
     def on_texcmd_change(self):
```
TheZoq2, line 440: The changes in this file seem to make these 2 functions redundant, should I just remove them?

textext/asktext.py
```diff
@@ -819,10 +818,7 @@ def cb_cancel(self, widget=None, data=None):
         self.window_deleted_cb(widget, None, None)
 
     def cb_compiler_changed(self, combo_box):
-        using_tex = self.TEX_COMMANDS[self._texcmd_cbox.get_active()] != "typst"
-        self._preview_button.set_sensitive(using_tex)
-        self._preamble_widget.set_sensitive(using_tex)
-        self._preamble_delete_btn.set_sensitive(using_tex)
+        pass
```
jcwinkler, line 821: Did you test the preview for typst?

textext/base.py
```diff
@@ -599,27 +599,19 @@ def typ_to_any(self, typst_command, typst_text, preamble_file, file_type):
...
-            # texwrapper = self.DOCUMENT_TEMPLATE % (preamble, latex_text)
-
+            # Read preamble
+            preamble_file = os.path.abspath(preamble_file)
+            preamble = ""
+            
+            if os.path.isfile(preamble_file):
+                with open(preamble_file, 'r') as f:
+                    preamble += f.read()
+
```
jcwinkler, line 609: As far as I do see this breaks the option to not having a document class in the latex preamble. This is a feature some users have so we introduced it.

jcwinkler (commented):
Thank you very much for your contribution to TexText. I clearly see that we need preamble support for typst. Hence, this feature should be implemented.

However, although your approach might work in your specific setup, it is still to narrow to capture all requirements we have to offer. 

For example: If we switch the compiler from pdflatex to typst: The preamble should switch from a tex file to a typ file. And if so, should we take a default file? Or a file which is stored in a node? Other things are broken, see my comments in the code.

I already implemented this for TexText 2.0 (where I am stuck in other issues) and I will backport it for the next release of TexText 1.x. This should be ready next week. I hope you are fine with this.

textext/asktext.py
TheZoq2: Yep

textext/base.py
TheZoq2: This is in `typ_to_any` so I figured it'd only affect typst side?

TheZoq2 (commented):
> I already implemented this for TexText 2.0 (where I am stuck in other issues) and I will backport it for the next release of TexText 1.x. This should be ready next week. I hope you are fine with this.

Oh cool! I'll just use my fork until then, thanks :)

jcwinkler:
@TheZoq2 I implemented preamble support in a more general manner and also activated the preview (see commits mentioning this PR). I added you to the list of authors since the key ideas are your ones. Thank you very much!

PR #456, opened 2025-03-04
Fix test on Linux

user202729:
On my machine, `pytest` fails because the `iec.which("inkscape")` raises `CommandNotFound` (subclass of `OSError`). This pull request fixes the issue by catching it.

Short checklist:
- [x] Tested with Inkscape version: 1.4

### DIFF
--- a/textext/requirements_check.py
+++ b/textext/requirements_check.py
@@ -624,19 +624,35 @@ class TexTextRequirementsChecker(object):
 
         return RequirementCheckResult(True, ["TkInter is found"])
 
+    def _find_inkscape_without_inkex(self):
+        """
+        Returns either (stdout_line, executable) or None
+        If the first is returned, stdout_line is the output of
+        inkscape --version and executable is the path to inkscape
+        """
+        try:
+            executable = self.find_executable('inkscape')['path']
+            stdout, stderr = defaults.call_command([executable, "--version"])
+            stdout_line = stdout.decode("utf-8", 'ignore')
+            return stdout_line, executable
+        except (KeyError, OSError):
+            return None
+
     def find_inkscape_1_4(self):
         try:
-            # When we call this from Inkscape we need this call
             import inkex.command as iec
-            stdout_line = iec.inkscape("", version=True)
-            executable = iec.which("inkscape")
         except ImportError:
+            result = self._find_inkscape_without_inkex()
+        else:
             try:
-                executable = self.find_executable('inkscape')['path']
-                stdout, stderr = defaults.call_command([executable, "--version"])
-                stdout_line = stdout.decode("utf-8", 'ignore')
-            except (KeyError, OSError):
-                return RequirementCheckResult(False, ["inkscape is not found"])
+                result = iec.inkscape("", version=True), iec.which("inkscape")
+            except iec.CommandNotFound:
+                result = None
+
+        if result is None:
+            return RequirementCheckResult(False, ["inkscape is not found"])
+        else:
+            stdout_line, executable = result
 
         m = re.search(r"Inkscape ((\d+)\.(\d+)[-\w]*)", stdout_line)
         if m:


jcwinkler:
I cannot reproduce this behaviour on my machine and also when the tests are run everything is fine. Could you investigate a little bit deeper, why the `which` command of the inkex module does not work?

What is the result of `inkex.command.is_inkscape_available()` and `inkex.command.inkscape("", "--version")` ?

user202729:
> I cannot reproduce this behaviour on my machine and also when the tests are run everything is fine. Could you investigate a little bit deeper, why the `which` command of the inkex module does not work?
> 
> What is the result of `inkex.command.is_inkscape_available()` and `inkex.command.inkscape("", "--version")` ?

It's specifically on the test that tests the situation where inkscape is not available. Maybe the difference is that in my case I have `inkex` installed globally while you don't.

```
_______________________________________________________________________________________ ERROR collecting test_installation_script.py _______________________________________________________________________________________
test_installation_script.py:94: in <module>
    test_configuration(bad_configuration, REQUIREMENT_CHECK_ERROR)
test_installation_script.py:68: in test_configuration
    assert ret_code == expected_exit_code, '%d != %d' % (ret_code, expected_exit_code)
E   AssertionError: 1 != 65
E   assert 1 == 65
```

…

→

```

Traceback (most recent call last):
  File ".../textext/setup.py", line 384, in <module>
    check_result = checker.check()
  File ".../textext/textext/requirements_check.py", line 765, in check
    check_result = textext_requirements.check()
  File ".../textext/textext/requirements_check.py", line 483, in check
    result = self.criteria()
  File ".../textext/textext/requirements_check.py", line 473, in <lambda>
    self.criteria = lambda: criteria(*args, **kwargs)
                            ~~~~~~~~^^^^^^^^^^^^^^^^^
  File ".../textext/textext/requirements_check.py", line 534, in and_impl
    L = self.check()
  File ".../textext/textext/requirements_check.py", line 483, in check
    result = self.criteria()
  File ".../textext/textext/requirements_check.py", line 473, in <lambda>
    self.criteria = lambda: criteria(*args, **kwargs)
                            ~~~~~~~~^^^^^^^^^^^^^^^^^
  File ".../textext/textext/requirements_check.py", line 534, in and_impl
    L = self.check()
  File ".../textext/textext/requirements_check.py", line 483, in check
    result = self.criteria()
  File ".../textext/textext/requirements_check.py", line 473, in <lambda>
    self.criteria = lambda: criteria(*args, **kwargs)
                            ~~~~~~~~^^^^^^^^^^^^^^^^^
  File ".../textext/textext/requirements_check.py", line 631, in find_inkscape_1_4
    stdout_line = iec.inkscape("", version=True)
  File ".../.local/lib/python3.13/site-packages/inkex/command.py", line 294, in inkscape
    stdout = call(INKSCAPE_EXECUTABLE_NAME, svg_file, *args, **kwargs)
  File ".../.local/lib/python3.13/site-packages/inkex/command.py", line 260, in call
    stdout = _call(program, *args, **kwargs)
  File ".../.local/lib/python3.13/site-packages/inkex/command.py", line 222, in _call
    args = to_args(which(program), *args, **kwargs)
                   ~~~~~^^^^^^^^^
  File ".../.local/lib/python3.13/site-packages/inkex/command.py", line 128, in which
    raise CommandNotFound(f"Can not find the command: '{program}'")
inkex.command.CommandNotFound: Can not find the command: 'inkscape'
```

jcwinkler:
Hmm, but either inkex is installed globally or not it should not make a difference. Is it maybe an `inkex` issue?. I gues your `inkscape` executable is in the path?

user202729:
> Hmm, but either inkex is installed globally or not it should not make a difference.

It should make a difference right? From my reading of the code,

* if inkex *is* installed globally then `import inkex.command as iec` would not raise `ImportError`, and `iec.which` is used (the problem is *in this case* if it raises `CommandNotFound` then it won't be caught);
* otherwise, `self.find_executable` is used.



> I gues your `inkscape` executable is in the path?

Yes, but it doesn't matter. The test

```
test_configuration(bad_configuration, REQUIREMENT_CHECK_ERROR)
```

specifically wipes the normal value of PATH and replace it with the temporarily directory.

jcwinkler:
I am talking from the view of the inkex module, not from the view of the requirements checker. It is clear that the requirements checker fails if `which` fails and the checker does not handle this properly. But what happens if you import `inkex` from a Python console. Does `which` for inkscape fails there also?

user202729:
No, from outside the test `which` works.

```
[textext]$ ipython
Python 3.13.1 (main, Dec  4 2024, 18:05:56) [GCC 14.2.1 20240910]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.31.0.dev -- An enhanced Interactive Python. Type '?' for help.

In [1]: import inkex

In [2]: inkex.command.which("inkscape")
Out[2]: '/usr/bin/inkscape'

In [3]: inkex.command.which("abcd")
---------------------------------------------------------------------------
CommandNotFound                           Traceback (most recent call last)
...
CommandNotFound: Can not find the command: 'abcd'
```

The reason why it fails inside the test is because the test is specifically testing the situation where inkscape is not available.

jcwinkler:
Now I got it, thank you @user202729 !

textext/requirements_check.py
```diff
@@ -624,19 +624,35 @@ def find_tkinter(self):
...
         except ImportError:
+            result = self._find_inkscape_without_inkex()
+        else:
             try:
-                executable = self.find_executable('inkscape')['path']
-                stdout, stderr = defaults.call_command([executable, "--version"])
-                stdout_line = stdout.decode("utf-8", 'ignore')
-            except (KeyError, OSError):
-                return RequirementCheckResult(False, ["inkscape is not found"])
+                result = iec.inkscape("", version=True), iec.which("inkscape")
```
jcwinkler, line 648: Under Windows, we have a problem here under special circumstances. While `import inkex.command` succeeds `iec.inkscape` may fail if `inkscape.exe` is not in the path. In that case we need to fall back to `_find_inkscape_without_inkex` again. I fixed this in the old approach by commit 58f846b (`CommandError` derives from `IOError`). 

We need to make this smooth without repeated try / except blocks.

jcwinkler (commented):
There is a small issue under Windows, please see my comments.

textext/requirements_check.py
user202729: I suppose that approach works too.

Issue #452, opened 2025-03-04
Speed up dialog creation by running a daemon


user202729:
There was a feature request in https://github.com/textext/textext/issues/263 about speeding up the creation.

Of course window creation and importing all the Python modules etc. take time, however this can be sped up by running a daemon persistently in the background, and each extension invocation is just a thin client connecting to the server/daemon.

Since the daemon has already imported everything in advance, there is no time wasted to imports.

jcwinkler:
In my opinion this is far beyond the scope of this extension. Please note that this will introduce a lot of overhead which, in turn, will be prone to errors. We also have to provide this on all three operating systems (Linux. MacOS, Windows). In my long time experience with TexText support takes a lot of time although we only have a single requirement: A working LaTex distro. But there are plenty of things that can go wrong: Python configuration issues on the OS, problems with the GTK3 framework etc.

user202729:
That's true. "Feature nice to have but too much work to support properly on all operating systems".

I suppose if someone (possibly me) propose to implement this later, you can just hide it behind some "experimental and unsupported, for power users only, might break any time" flag?

jcwinkler:
If it does not break the standard behaviour I have no problem with that.

PR #457, opened 2025-03-04
Show traceback in error dialog

user202729:
As in the title. This makes debugging easier if some error happens.

By the way, what is the https://github.com/textext/textext/tree/TEXTEXT_2_0_Dev branch?

Short checklist:
- [ ] Tested with Inkscape version: [fill in Inkscape version here]
- [ ] Tested on Linux, Distro: [fill in distribution name here]

### DIFF
--- a/textext/asktext.py
+++ b/textext/asktext.py
@@ -38,6 +38,7 @@ TOOLKIT = None
 import os
 import sys
 import warnings
+import traceback
 from .errors import TexTextCommandFailed
 from textext.utility import SuppressStream
 
@@ -509,6 +510,8 @@ class AskTextTK(AskText):
 
             if exception.stderr:
                 add_textview('Stderr:', exception.stderr.decode('utf-8'))
+
+        add_textview('Traceback:', '\n'.join(traceback.format_exception(exception)))
 
         close_button = Tk.Button(err_dialog, text='OK', command=close_error_dialog)
         close_button.pack(side='top', fill='x', expand=True)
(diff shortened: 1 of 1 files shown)


jcwinkler:
Thanks, I will test it.

jcwinkler:
> By the way, what is the https://github.com/textext/textext/tree/TEXTEXT_2_0_Dev branch?

Currently, I am re-writing the extension from scratch since the code is, as you will have experienced, completey weird after 15+ years of development. From time to time I push the current state of the development into this branch (it is not up-to-date currently). Honestly, progress is very slow since I have to do a lot of other things. But there is progress.

Let me know if you are willing to contribute more on a regular basis. Then I will update the branch. It is heavily WIP. Please do not waste too much time with the old code you find in `develop`.

user202729:
> Thanks, I will test it.

Thanks. (I've only tested one of the GUI (GTK I think), not the other one. )

user202729:
> the code is, as you will have experienced, completey weird after 15+ years of development

Actually it seems just fine (I've seen worse). And… it does the job, so why not.

> Let me know if you are willing to contribute more on a regular basis. Then I will update the branch.

Honestly… maybe not really, unless there's some feature that I need. But seeing the code would be great.

On the other hand I think a drawback of contributing to that branch instead of 1.11, that way there are less users being able to see the outcome.

user202729:
By the way, if I consider to implement something would you prefer if I target it against 1.11 or 2.0? Although I haven't seen the latest 2.0 code, you said it doesn't have as many features, which would be another downside? (Or is it not that bad and/or have some other important features?)

jcwinkler:
2.0 will have more features compared to 1.x. If it is urgent for you to have the features you mentioned you may implement them in 1.x (-> develop). We could make a 1.12 release and we can port the features later to 2.0. I guess I will merge the new code within the next 2 or 3 weeks. Maybe I will branch off a 1.x branch or sth. like this. I have to think carefully about this the next days.

jcwinkler:
@user202729 Please provide any input into develop, i.e. to 1.x. I suggest you are most interested in #452?

PR #461, opened 2025-03-17
Unique IDs also in children

### LINKED ISSUE #460: Individual character being selected after re-edit an element
1. create an TexText element with content let's say "12"
2. edit it to content "34"
3. after you're done notice that only the "3" is selected instead of the whole thing

https://github.com/user-attachments/assets/c58417d5-ad6c-4728-b646-5f829e579c91

(at least the group is not entered, which is good I guess)

jcwinkler:
Using inkex API to set unique IDs for the node and also its children. Ensures that
the old node is still selected completely after TexText closes, see #460.

Resolves #460

@user202729 I think this fixes the problem. If you have the time, you may test it.

Short checklist:
- [x] Tested with Inkscape version: 1.4.0
- [x] Tested on Linux, Kubuntu 24.04
- [x] Tested on Windows, Version: 11 23H2

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -744,32 +744,10 @@ class TexTextElement(inkex.Group):
     def make_ids_unique(self):
         """
         PDF->SVG converters tend to use same ids.
-        To avoid confusion between objects with same id from two or more TexText objects we replace auto-generated
-        ids with random unique values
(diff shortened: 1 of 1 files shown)


user202729:
Looks like it's working, thanks.

Issue #453, opened 2025-03-04
Recompile all object/modify preamble from command-line


user202729:
There is https://github.com/textext/textext/issues/201 which suggests having a button in the GUI to recompile all objects.

However, often in a document there are multiple SVG files in a directory, and opening inkscape for each of them would be slow.

Instead, it would be nice if we can have some way to modify preamble and recompile all objects from the command-line. That way it would be easy to script it to loop over all files in a directory.

jcwinkler:
Thanks @user202729 . So from a UI perspective I see two scenarios:

- A user changed something in the preamble such that it affects the outcome of the compilation. Hence, the user wants to recompile all nodes but does not want to open each snippet separately to do so
- A user explicitly changed the scale of one node. Now the user wants to change the scale of all other nodes as well.

We could add a button: `Recompile all TexText nodes` to the UI (covers the preamble case)

We could add another button `Recompile all TexText nodes with this scale factor` (covers the scale factor case)

Do you agree? Do you see other scenarios?

(How the buttons are labeled and where they are placed can be discussed later)

Would you like to open a pull-request (maybe as WIP) ?

user202729:
> I think a lot of users (including me) just draw something and add the text later leading to scale factors different from 1.0. Hence, it would be convenient if TexText offers the possibility to re-scale all nodes at once.

The reason for me is because I include the figures in a TeX file, and I care about typography details so I want to font-match the figures with the document. My current workflow is to never scale either the figures or the font, and that works out very well.

> Progress dialog for users during batch compile (could be take a lot of time)

Sure that's nice, but there's a way to speed this up. Easier to explain in code than in words though (basically use a series of `preview` environment, one figure per page, then extract out the pages later, so you only need one LaTeX compilation. Error handling considerations aside)

PR #468, opened 2025-06-10
Recompile all entries on command-line

user202729:
Partial progress towards https://github.com/textext/textext/issues/453 . 

The explanation how to use the feature can be found in docstring

        Mutate ``self.svg`` to recompile all textext entries.
        This can be invoked from command-line as::
            python3 /path/to/textext/__main__.py --recompile-all-entries        > edited.svg < original.svg
            python3 /path/to/textext/__main__.py --recompile-all-entries --output edited.svg < original.svg
        In the first form ``edited.svg`` must not be the same as ``original.svg``,
        in the second form it is probably fine (although do make a backup).

I did some manual testing. First create `/tmp/preamble.tex` (attached below), then create `/tmp/a.svg` (also attached below), then modify `preamble.tex` to 

```
\usepackage{amsmath,amssymb,newtxmath}
```

then run the command, and see that the output's font correctly changes.

[a.zip](https://github.com/user-attachments/files/20679472/a.zip)

Short checklist:
- [x] Tested with Inkscape version: 1.4.1

### DIFF
--- a/textext/base.py
+++ b/textext/base.py
@@ -334,6 +334,100 @@ class TexText(inkex.EffectExtension):
                     converter.pdf_to_png(white_bg=white_bg)
                     image_setter(converter.tmp('png'))
 
+    def _do_convert_one(self, text: str, preamble_file, user_scale_factor, alignment, tex_command):
+        """
+        Does the conversion using the selected converter.
+        See documentation in do_convert for more details.
+        """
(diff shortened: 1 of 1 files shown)


user202729:
so what was the problem on your environment? Users might encounter it too.

jcwinkler:
No, it was just a configuration problem of the Debug-environment in PyCharm.

I will add a new section "Command Line Usage" to the documentation now (which is also interesting for all the testing stuff).

PR #471, opened 2025-06-23
Allow the user to select GUI toolkit

user202729:
Allow the user to configure the GUI toolkit being used, either GTK or TK.

The motivation is to make testing (whether some change works in both TK and GTK) in development easier. Previously I think in order to test TK you need to uninstall GTK.

Currently there's no GUI to configure it yet (since there are only two I'm thinking making just a single button "switch to GTK GUI" / "switch to TK GUI"). I manually edit `config.json` in `~/.config/textext/` (on Linux) and add e.g.

```
{
  "preamble": ...
  "scale": 1.0,
  "previous_tex_command": "pdflatex",
  "gui": {
    "toolkit": "tk",  // << add this
    "confirm_close": true
  }
}
```

Because currently there's no GUI, I think raising hard error (instead of handle it gracefully) is acceptable. Feature to e.g. automatically fallback to the other GUI and warn the user can be implemented later.

The difference between TK and GTK is clear, but I see no difference between GTK and GTK SourceView. I'm not sure where to look for exactly.

The diff generated by git is huge, but the "semantic diff" is small:

* move the classes to the function (add indentation) (the indentation throws git diff off)
* add initial assignment "GtkSource = None"
* change "TOOLKIT == GTKSOURCEVIEW" to "GtkSource is not None"
* actually use WINDOW_TITLE
* move the detection code to base.py

It may make reviewing easier by using `--ignore-all-spaces` flag in `git diff`.

For some reason when an error happens early on (incorrect toolkit) then the tool will fail silently. As a workaround I add this

```diff
diff --git a/textext/__main__.py b/textext/__main__.py
index 05a4a93..81d0c06 100644
--- a/textext/__main__.py
+++ b/textext/__main__.py
@@ -54,6 +54,7 @@ if __name__ == "__main__":
     except Exception as e:
         # All errors should be handled by above clause.
         # If any propagates here it's TexText logic error and should be reported.
+        sys.stderr.write(str(e) + '\n')
         logger.error(str(e))
         logger.error(traceback.format_exc())
         logger.info("TexText finished with error, please run extension again")
```

to make the error message visible to the user.

Short checklist:
- [x] Tested with Inkscape version: 1.4.1

### DIFF
--- a/textext/asktext.py
+++ b/textext/asktext.py
@@ -29,95 +29,12 @@ $"""
 
 WINDOW_TITLE = "Enter LaTeX Formula - TexText"
 
-GTKSOURCEVIEW = "GTK Source View"
-GTK = "GTK"
-TK = "TK"
-
-TOOLKIT = None
-
 import os
 import sys
 import warnings
 import traceback
 from .errors import TexTextCommandFailed
 from textext.utility import SuppressStream
-
-# unfortunately, with Inkscape being 32bit on OSX, I couldn't get GTKSourceView to work, yet
-
-# Try GTK first
-#   If successful, try GTKSourceView (bonus points!)
-#   If unsuccessful, try TK (first for Python 3, then for Python 2)
-#   When not even TK could be imported, abort with error message
-try:
-    import gi
-    gi.require_version("Gtk", "3.0")
-
-    # The import statement
-    # from gi.repository import Gtk
-    # writes a warning into stderr under Python 3.10 which always pops up after the
-    # extensions has been executed:
-    # "DynamicImporter.exec_module() not found; falling back to load_module()"
-    # We redirect stderr here into a string, check if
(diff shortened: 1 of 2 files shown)


jcwinkler:
Thanks! Some remarks:

> The motivation is to make testing (whether some change works in both TK and GTK) in development easier. Previously I think in order to test TK you need to uninstall GTK.

I did testing of TK by just changing `import gi` to `import gix` which hands over to tkinter. Not nice, I know. I like your approach, it will especially help MacOS users. There, often the import of the gobject framework succeeds but when calling some methods it crashes. The user does not have a chance to go into tkinter as a fallback without changing the source code. By your approach this can be achieved by modifying the config file.

I suggest to add an option `--force-tkinter` in the setup script so the config file can be properly setup during installation.

PR #485, opened 2026-01-05
Switch to GTKSourceView 4

### LINKED ISSUE #483: TexText not working in Inkscape 1.43 (Windows)
Hi
The Tex Tex dialog doesn't show up anymore after updating Inkscape from 1.42 to 1.43.

It was working fine on Inkscape 1.42.
No error message. The dialog for typing equations doesn't show up.

Thanks for this wonderful plugin!

**System information**

- TexText version: 1.12
- Inkscape version: 1.43
- Operating system: Windows 11 x64
- Windows only: LaTeX-distribution MiKTeX
- Windows only: TexText installed via installer exe

jcwinkler:
GtkSourceView is now the default in Inkscape >= 1.43 and also packaged in Windows. Hence, we use it now instead of GTKSourceView 3.0.

Resolves #483 

Short checklist:
- [x] Tested with Inkscape version: 1.43
- [x] Tested on Linux, Distro: Ubuntu 24.04
- [x] Tested on Windows, Version: 11 23H2


jcwinkler:
CI test fails since they still use Inkscape 1.4.2 so the installation script fails. Hence, I merge it now. All tests passed locally.

Issue #448, opened 2025-02-15
Export with font matching?


user202729:
**Is your feature request related to a problem? Please describe.**
When an svg is created with textext and it's embedded into a LaTeX document, and the font of the outer LaTeX document changes, the font in the figures does not change accordingly.

**Describe the solution you'd like**
One way out is the following: inkscape has a feature to "omit text in pdf and create latex file", see https://castel.dev/post/lecture-notes-2/ . However in order to create such a text file, it is necessary to use the raw text feature in inkscape.

It would be good if there's some way (possibly a button on the plugin) to export pdf+latex of the current document after converting all textext groups to normal text box.

jcwinkler:
Hmm, we have to think about this in more detail. Currently we have: LaTeX-Code -> PDF file -> SVG snippets with all symbols converted to paths.

If we change the behaviour of TexText such that it does export the raw text what would be the use of TexText? Why not entering the text directly in Inkscape as given in the example you linked?

Could you explain it in more detail, please. Or give an example svg showing what you would like to have exported as text and waht you would like to have compiled by TexText?

user202729:
In that case, one advantage of TexText over entering the text directly in Inkscape would be to directly preview where the rendered TeX would be placed on the canvas.

As mentioned in https://castel.dev/post/lecture-notes-2/ : using raw text followed by import to pdf+latex has the **advantage of font matching**, but **disadvantage of being difficult to position the text correctly**.

jcwinkler:
OK, please check if I got it correctly:

- TexText works as usual. all nodes are compiled and the rendered output is placed into the SVG
- There exists a button in TextText. When it is pressed the SVG is taken, all rendered TexText output is replaced by the LaTeX code and this document is exported as "Omit text in PDF and create LaTeX file".

This should be possible to implement. Before doing this it would be necessary to check how the alignment is done with a snippet like `$x = \sqrt{y}$`. Will the rendered code (in the tex document) be aligned to the top left corner of the original tex code? Or centered?. Maybe this sketch illustrates my question:

![Image](https://github.com/user-attachments/assets/68faf405-ad62-4ec3-8709-07b5f600a72b)

Could you check this?

user202729:
Yes, sounds correct.

As for the alignment… looks like it is partially determined by the text alignment setting.

![Image](https://github.com/user-attachments/assets/0b7c0c0b-f7dd-40e9-a1fa-7acdc2be5892)

[...]

PR #491, opened 2026-02-05
[WIP] Feature: export pdf latex

Frischid:
Related issue(s): #448

Based on previous work of @user202729 (some comments are from the original author and might need to be checked if still valid) with added scaling meta and proper vertical alignment without clipping. Allows to export as pdf_tex file with the textext elements converted back to LaTeX code with proper positioning. This should allows to e.g. match the font of the parent LaTeX, hyperlinks with \href, \cite,...

In order to test it, you can run the following lines:
``` bash
# recalculate offsets
python3 textext/__main__.py --recompile-all > edited.svg < Test.svg
# export as pdf_tex to currently /tmp/a*
python3 textext/__main__.py --export-pdf-latex edited.svg
# calculate the different pixels
# although the number might look high with 1000 dpi, the visible differences are minimal and might be due to rounding
diff-pdf -v /tmp/a_reference.pdf /tmp/a_preview.pdf --dpi=1000 --output-diff=out.pdf | head -n1

# to visually inspect run the following or open out.pdf from previous command:
diff-pdf --view /tmp/a_reference.pdf /tmp/a_preview.pdf --dpi=1000

```
Short checklist:
- [x] Tested with Inkscape version: Inkscape 1.4.3 (0d15f75042, 2025-12-25)
- [x] Tested on Linux, Distro: Manjaro/ArchLinux

Todo:
- [x] Fix clipping/offsets
- [x] Fix scaling
- [ ] Improve export instead of writing to `/tmp`
- [ ] Test with more Distros and Testfiles

### DIFF
--- a/textext/asktext.py
+++ b/textext/asktext.py
@@ -102,7 +102,7 @@ class AskText(object):
         self._cancel_button = None
         self._window = None
 
-    def ask(self, callback, preview_callback=None):
+    def ask(self, callback, preview_callback=None, export_with_font_matching_callback=None):
         """
         Present the GUI for entering LaTeX code and setting some options
         :param callback: A callback function (basically, what to do with the values from the GUI)
@@ -203,7 +203,7 @@ def load_asktext_tk():
                     valid = False
             return valid
 
-        def ask(self, callback, preview_callback=None):
+        def ask(self, callback, preview_callback=None, export_with_font_matching_callback=None):
             self.callback = callback
 
             self._root = Tk.Tk()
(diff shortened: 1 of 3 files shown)


Frischid:
I was trying to figure out the best way to do the export since I guess we don't know the input filename.
Would it make sense to use an OutputExtension? https://inkscape.gitlab.io/extensions/documentation/source/inkex.extensions.html?highlight=effectextension#inkex.extensions.OutputExtension
Or providing a command-line option and gui input for the final filename for .pdf and .pdf_tex files?
The replacement and alignment of textext seems to be fine now.

